PR #77, opened 2020-04-01
Trivy get top level owner

baderbuddy:
Due to a change in the logic we started pulling the owner of pods, but don't follow the chain all the way up. This PR changes the logic. It also introduces a need for more permissions for the trivy service account, so a chart change will be needed too.


trivy/pkg/getimages.go
```diff
@@ -62,20 +78,36 @@ func GetImages() ([]Image, error) {
...
-			owner.Name = pod.ObjectMeta.Name
-		}
-		key := owner.Namespace + "/" + owner.Kind + "/" + owner.Name
-		if _, ok := found[key]; ok {
-			continue
+		owners := pod.ObjectMeta.OwnerReferences
+
+		for len(owners) > 0 {
+			if len(owners) > 1 {
+				logrus.Warn("More than 1 owner found")
```
rbren, line 88: can you add namespace/name to this message?

trivy/pkg/getimages.go
```diff
@@ -62,20 +78,36 @@ func GetImages() ([]Image, error) {
...
+			}
+			firstOwner := owners[0]
+			owner.Kind = firstOwner.Kind
+			owner.Name = firstOwner.Name
+			fqKind := schema.FromAPIVersionAndKind(firstOwner.APIVersion, firstOwner.Kind)
+			mapping, err := restMapper.RESTMapping(fqKind.GroupKind(), fqKind.Version)
+			if owner.Kind == "Node" {
+				break
```
rbren, line 96: Is this a thing? What workloads are owned by a `Node`?

PR #81, opened 2020-04-06
Turn off DLC

rbren:


.circleci/config.yml
```diff
@@ -36,8 +36,6 @@ jobs:
       - image: quay.io/reactiveops/ci-images:v11-stretch
     steps:
       - checkout
-      - setup_remote_docker:
```
baderbuddy, line 39: Oh we still need the remote docker line, just make the layer caching flase

PR #97, opened 2020-04-29
Fix Kubesec

baderbuddy:
A couple problems, the item file wasn't being overwritten so we always got the same result back and exit codes were crashing out, so any insecure pods would crash kubesec.


rbren:
Oh wow. Thanks for fixing!

PR #98, opened 2020-04-29
Increase trivy version number

baderbuddy:


rbren:
API/output is unchanged I assume?

PR #121, opened 2020-05-28
Initial code for CI plugin

baderbuddy:
Refactored the Trivy code to make it more reusable and then wrote some initial code for the CI utility.


ci/README.md
```diff
@@ -0,0 +1,3 @@
+# ci
+
+A utility for the CI/CD integration of Fairwinds Insights.
```
rbren, line 3: Mind adding some usage instructions here?

ci/pkg/ci/ci.go
```diff
@@ -0,0 +1,306 @@
...
+				allRepoTags = append(allRepoTags, tag.(string))
+			}
+		}
+		return allRepoTags, nil
+	}
+	return nil, nil
+}
+
+// SendResults sends the results to Insights
+func SendResults(trivyResults []byte, trivyVersion string, polarisVersion string, configurationObject Configuration, token string) error {
```
rbren, line 160: We'll probably want to do something that scales as we add more reports - maybe an array of `Result` structs?

I'm guessing most tools will output to a file, so maybe it makes sense to just write out the trivy results so that we can have a consistent method for accessing results?

ci/pkg/ci/ci.go
```diff
@@ -0,0 +1,306 @@
...
+		return err
+	}
+
+	branchName, err := GetResultsFromCommand("git", "rev-parse", "--abbrev-ref", "HEAD")
+	if err != nil {
+		logrus.Warn("Unable to get GIT Branch Name")
+		return err
+	}
+
+	origin, err := GetResultsFromCommand("git", "remote", "get-url", "origin")
```
rbren, line 214: might this contain an access key? should we remove pattern `://([^@]+@)`?

baderbuddy: I'm not positive, but better safe than sorry.

baderbuddy:
@rbren I made some changes based on your feedback. I feel like this lead to some code that still needs to be cleaned up. But especially the Filename issue I wanted to see what you thought of what I had here. While I'm scanning the YAML files I capture the comment and filename of each resource and match any Action Items that come back to those to figure out what file they came from.

ci/cmd/insights-ci/main.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// Scan manifests with Polaris
+	polarisReport, err := getPolarisReport(configurationObject)
+	if err != nil {
+		panic(err)
+	}
+	// Send Results up
```
rbren, line 58: [ahh comments, my old nemesis](https://blog.humphd.org/vocamus-894/)

ci/cmd/insights-ci/main.go
```diff
@@ -0,0 +1,156 @@
...
+			return trivyReport, err
+		}
+		images[idx].PullRef = strconv.Itoa(idx)
+	}
+	// Scan Images with Trivy
+	trivyResults, trivyVersion, err := ci.ScanImagesWithTrivy(images, configurationObject)
+	if err != nil {
+		return trivyReport, err
+	}
+	err = ioutil.WriteFile(configurationObject.Options.TempFolder+"/trivy.json", trivyResults, 0644)
```
rbren, line 131: should this be `+ trivyReport.Filename`?

ci/pkg/ci/ci.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	return strings.TrimSpace(string(bytes)), err
+}
+
+// GetImagesFromManifest scans a folder of yaml files and returns all of the images used.
+func GetImagesFromManifest(configFolder string) ([]models.Image, []Resource, error) {
+	images := make([]models.Image, 0)
+	resources := make([]Resource, 0)
+	err := filepath.Walk(configFolder, func(path string, info os.FileInfo, err error) error {
+		if strings.HasSuffix(info.Name(), ".yaml") {
```
rbren, line 40: since this gets so indented, I'd suggest using a `continue` here instead

ci/pkg/ci/ci.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		logrus.Warn("Unable to get GIT Origin")
+		return err
+	}
+
+	if strings.LastIndex(origin, "@") > 3 { // git@github.com URLs are allowed
+		logrus.Warn("Detected possible token in Git Origin, stripping git origin.")
+		originSplit := strings.Split(origin, "@")
+		// Take the substring after the last @ to avoid any tokens in an HTTPS URL
+		origin = originSplit[len(originSplit)-1]
```
rbren, line 233: this will generate inconsistent origin URLs, since those that don't contain an `@` will still have the protocol

ci/pkg/ci/ci.go
```diff
@@ -0,0 +1,327 @@
...
+		if err != nil {
+			return err
+		}
+		xmlBytes = append([]byte(xml.Header), xmlBytes...)
+		err = ioutil.WriteFile(configurationObject.Options.JUnitOutput, xmlBytes, 0644)
+		if err != nil {
+			return err
+		}
+	}
+	logrus.Infof("Score of %f with a baseline of %f", results.Score, results.BaselineScore)
```
rbren, line 320: I'd suggest breaking out the results handling (and maybe the junit output) from the HTTP logic

ci/cmd/insights-ci/main.go
baderbuddy: Those comments were really pretty originally!

PR #138, opened 2020-06-11
Change format of CI Config

baderbuddy:
New format for the config file, new workloads report, moving away from Score


baderbuddy:
I forgot one other change to make to this is changing the logic for how the base hash is picked.

PR #149, opened 2020-06-23
Send error reports on failure

baderbuddy:
This won't work yet, it'll require permissions changes in the chart as well as changes on the Insights server end to go into prod first. But this should query for failures in the pod that the insights-uploader is running in and send those logs up to Insights if there's a failure.


baderbuddy:
I added the environment variable check to not send the logs if disabled in the Helm chart.

PR #167, opened 2020-07-14
Fix up uploader script

rbren:


rbren:
Oops, forgot to bump the version!

PR #184, opened 2020-07-20
Calculate pod count

baderbuddy:


workload/pkg/info.go
```diff
@@ -224,6 +218,146 @@ func CreateResourceProviderFromAPI(ctx context.Context, kube kubernetes.Interfac
...
+		}
+		ownerReferences := item.GetObjectMeta().GetOwnerReferences()
+		ownerUID := getOwnerUID(ownerReferences)
+
+		if ownerUID != "" {
+			ownerController := controllerMap[ownerUID]
+			if ownerController != nil {
+				controllerMap[ownerUID].PodCount++
+				ownerContainers := controllerMap[ownerUID].Containers
+				if len(ownerContainers) == 0 || (len(containers) > 0 && containers[0].CreationTime.After(ownerContainers[0].CreationTime)) {
```
rbren, line 307: not sure I understand why we swap out ownerContainers for podContainers - when would they differ?

workload/pkg/info.go
```diff
@@ -224,6 +218,146 @@ func CreateResourceProviderFromAPI(ctx context.Context, kube kubernetes.Interfac
...
+
+	finalInterfaces := []ControllerResult{}
+	for id := range topController {
+		controller := controllerMap[id]
+		var count float64 = controller.PodCount
+		if controller.Kind != "CronJob" {
+			for _, childID := range children[id] {
+				count += controllerMap[childID].PodCount
+			}
+		} else {
```
rbren, line 329: I wonder if for `Job`, we should just do `duration / (now() - startTime)`...

workload/pkg/info.go
```diff
@@ -224,6 +218,146 @@ func CreateResourceProviderFromAPI(ctx context.Context, kube kubernetes.Interfac
...
+					}
+					if timeStamps.startTime.After(maxTime) {
+						maxTime = timeStamps.startTime
+					}
+				}
+				totalTime := maxTime.Sub(minTime).Seconds()
+				durationSum = (durationSum / float64(len(children))) * float64(len(children)-1)
+				count = durationSum / totalTime
+			} else {
+				count = 1
```
rbren, line 354: ditto on the `Job` calculation - maybe `duration / (now() - startTime)`? I think that's still an _over_ estimate in all cases...if the cj just ran, it'll give `1`, but if it only runs once/year, the pod count will slowly decay toward `0`.

workload/pkg/info.go
baderbuddy: The container status (including Image ID) isn't there at the Deployment level, so this tries to get the latest Pod and use that for container status.

workload/pkg/info.go
baderbuddy: I like it. Yeah this feels like a good way to get even more accurate numbers.

PR #185, opened 2020-07-28
CI Tweaks

rbren:
A couple changes here:
* created a separate `RunCommand` that doesn't have trivy-specific logic
* started logging stdout/stderr directly instead of through logrus, to preserve newlines
* fixed formatting on message
* remove `panic`s
* push helm values to file instead of using `--set`
* change output message
* set config defaults


rbren:
Gosh we need an automated check for that

PR #186, opened 2020-07-28
add version checking

rbren:
Let's see if this works


rbren:
Good call!

Sorry, clearly forgot about this.

PR #187, opened 2020-08-07
Initial add of OPA

baderbuddy:


opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/restmapper"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const outputFile = "/output/opa.json"
+
+func main() {
+	ctx := context.TODO()
```
rbren, line 27: I still don't quite get `context`, but from what I'm reading in the docs, `Background` is generally used by `main`, and `TODO` by functions that don't yet have a context being passed in

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+	instanceGvr := schema.GroupVersionResource{Group: "insights.fairwinds.com", Version: "v1", Resource: "customcheckinstances"}
```
rbren, line 28: should we start with `v1beta1`?

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/restmapper"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const outputFile = "/output/opa.json"
+
+func main() {
```
rbren, line 26: mind breaking `main` up a bit? Maybe something like `getKubeClient()`, `getAllActionItems()`, and `sendOutput()`

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		panic(err)
+	}
+	actionItems := make([]ActionItem, 0)
+
+	for _, checkInstance := range checkInstances.Items {
+		var checkInstanceObject customCheckInstance
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(checkInstance.Object, &checkInstanceObject)
+		if err != nil {
+			panic(err)
```
rbren, line 59: we should avoid panicking inside this loop, so we don't lose the whole report if there's a small issue

rbren: Optionally, when images fail to scan in Trivy, we send that back to the server, and create a single action item saying `12 images failed to scan`. Maybe helpful here if e.g. we don't have the right RBAC to access a resource they asked us to check

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+								}
+								strValue := str.String()
+								if len(strValue) > 0 && strValue[0] == '"' {
+									strValue = strValue[1:]
+								}
+								if len(strValue) > 0 && strValue[len(strValue)-1] == '"' {
+									strValue = strValue[:len(strValue)-1]
+								}
+								var apiGroup string
+								for _, target := range check.Spec.AdditionalKubernetesData {
```
rbren, line 127: we're getting pretty nested here 😅

Mind breaking out a separate function for the inside of this loop? Maybe the rego parts in particular.

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+
+func processResults(resource unstructured.Unstructured, results rego.ResultSet, check customCheck, checkInstance customCheckInstance) ([]ActionItem, error) {
+	actionItems := make([]ActionItem, 0)
+	instanceOutput := checkInstance.Spec.Output
+	checkOutput := check.Spec.Output
+	for _, result := range results {
+		for _, pack := range result.Bindings["results"].(map[string]interface{}) {
+			for _, outputArray := range pack.(map[string]interface{}) {
+				for _, output := range outputArray.([]interface{}) {
+					severity := instanceOutput.Severity
```
rbren, line 186: should this be `checkOutput`?

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+						category = instanceOutput.Category
+					}
+					strMethod, ok := output.(string)
+					var description string
+					if ok {
+						description = fmt.Sprintf("String: %s", strMethod)
+					} else {
+						mapMethod, ok := output.(map[string]interface{})
+						if ok {
+							description = fmt.Sprintf("Map: Desc: %v", mapMethod["description"])
```
rbren, line 209: not sure I understand what these descriptions are saying - can you give an example of what it might output?

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+							if mapMethod["severity"] != nil {
+								severityFloat, err := strconv.ParseFloat(mapMethod["severity"].(string), 64)
```
rbren, line 211: ditto on nesting - most of these lines are wrapping for me

kinda surprised gofmt doesn't have a max line length

opa/pkg/main.go
```diff
@@ -0,0 +1,263 @@
...
+								title = &titleString
+							}
+
+							if mapMethod["remediation"] != nil {
+								remediationString := mapMethod["remediation"].(string)
+								remediation = &remediationString
+
+							}
+						} else {
+							description = fmt.Sprintf("Could not decipher output format of %+v %T", output, output)
```
rbren, line 228: we should log this and just make the description blank. I think it'd be confusing to see this inside an action item

opa/pkg/main.go
baderbuddy: Ahh I was thinking that felt wrong. Thanks!

opa/pkg/main.go
baderbuddy: Yes it should

opa/pkg/main.go
baderbuddy: They were debugging things. The actual `mapMethod["description"]` is what should be set to description here. Which is just whatever was returned by the rego.

opa/pkg/main.go
baderbuddy: Yeah that nesting was getting out of control. I moved out the getKubernetesData function there

opa/pkg/main.go
baderbuddy: This feels like basically malformed rego. I'm going to treat this as an error.

opa/pkg/main.go
baderbuddy: This is the one piece of feedback here I'm struggling with. Let's just think through this. Errors are either temporary or permanent. If it's a temporary error then failing the report seems appropriate, we don't end up with any flop in Action items, they get their failure notification (assuming they didn't turn it off) and odds are in an hour everything will be fine.

If it's a permanent error (like not having the right RBAC or bad Rego). Have an action item to fix it could make sense. But it's also directly user error.

So if we fail the report, it punishes the user for their mistake and they're more likely to get it fixed right away.
If we create an action item for the failure, then we're being gracious towards the user and giving them as many results as possible.

Let me try to look through the panics and see which ones feel like we should be creating action items for them. I think at least some of these will end up that way.

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+			continue
+		}
+		var checkObject customCheck
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(check.Object, &checkObject)
+		if err != nil {
+			return nil, err
+		}
+
+		newItems, err := processCheck(ctx, checkObject, checkInstanceObject, restMapper, dynamicInterface)
+		if err != nil {
```
rbren, line 101: looks like we're hitting the k8s API in `processCheck` - probably best to `continue` here

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+		}
+		logrus.Infof("Starting to process check: %s", checkInstanceObject.Name)
+		check, err := dynamicInterface.Resource(checkGvr).Namespace(checkInstanceObject.Namespace).Get(ctx, checkInstanceObject.Spec.CustomCheckName, metav1.GetOptions{})
+		if err != nil {
+			actionItems = append(actionItems, ActionItem{
+				Title:       checkInstanceObject.Name,
+				Description: "An error occured retrieving the Custom Check for this instance",
+				Remediation: "Make sure that the custom check exists and it is in the same namespace as this instance.",
+				Severity:    0.4,
+				Category:    "Reliability",
```
rbren, line 89: Should we set `ResourceName` etc from the custom check?

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+		if !ok {
+			return nil, nil
+		}
+		strValue := str.String()
+		if len(strValue) > 0 && strValue[0] == '"' {
+			strValue = strValue[1:]
+		}
+		if len(strValue) > 0 && strValue[len(strValue)-1] == '"' {
+			strValue = strValue[:len(strValue)-1]
+		}
```
rbren, line 122: why the need to strip `"`?

`strings.Trim(strValue, "\"")` is probably the way to go though (unless internal `"` is somehow OK)

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	return actionItems, nil
+}
+
+func getKubernetesDataFunction(ctx context.Context, check customCheck, restMapper meta.RESTMapper, dynamicInterface dynamic.Interface) func(rego.BuiltinContext, *ast.Term) (*ast.Term, error) {
+	return func(_ rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
+
+		str, ok := a.Value.(ast.String)
+		if !ok {
+			return nil, nil
```
rbren, line 114: should we be logging something here? Returning an error?

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+			strValue = strValue[1:]
+		}
+		if len(strValue) > 0 && strValue[len(strValue)-1] == '"' {
+			strValue = strValue[:len(strValue)-1]
+		}
+		var apiGroup string
+		for _, target := range check.Spec.AdditionalKubernetesData {
+			fmt.Printf("Checking if %s matches %s", strValue, target.Kinds[0])
+			if strValue == target.Kinds[0] {
+				fmt.Printf("It does!")
```
rbren, line 127: debug? above too

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+
+}
+func processCheck(ctx context.Context, check customCheck, checkInstance customCheckInstance, restMapper meta.RESTMapper, dynamicInterface dynamic.Interface) ([]ActionItem, error) {
+	actionItems := make([]ActionItem, 0)
+
+	getKubernetesData := getKubernetesDataFunction(ctx, check, restMapper, dynamicInterface)
+	for _, target := range checkInstance.Spec.Targets {
+		for _, apiGroup := range target.APIGroups {
+			for _, kind := range target.Kinds {
+				logrus.Infof("Starting to process %s %s\n\n", apiGroup, kind)
```
rbren, line 161: this is still a little hard to read on GH - any chance we can break out the inner loop into its own function?

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+						return nil, err
+					}
+					newItems, err := processResults(obj, results, check, checkInstance)
+					if err != nil {
+						actionItems = append(actionItems, ActionItem{
+							Title:       checkInstance.Name,
+							Description: "An error occured processing the results of this check.",
+							Remediation: "Make sure that the return values are all correct.",
+							Severity:    0.4,
+							Category:    "Reliability",
```
rbren, line 216: Ditto on these AIs - they'll get deduped with eachother without a resource name/namespace/kind

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+func processResults(resource unstructured.Unstructured, results rego.ResultSet, check customCheck, checkInstance customCheckInstance) ([]ActionItem, error) {
+	actionItems := make([]ActionItem, 0)
+	instanceOutput := checkInstance.Spec.Output
+	checkOutput := check.Spec.Output
+	for _, output := range getOutputArray(results) {
+		severity := checkOutput.Severity
+		title := checkOutput.Title
+		remediation := checkOutput.Remediation
+		category := checkOutput.Category
+		if instanceOutput.Severity != nil {
```
rbren, line 253: For my understanding:
* `checkOutput` is the check definition
* `instanceOutput` is the check instance (which links the check to certain resources)
* `output` is what came out of running the rego

correct?

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+		strMethod, ok := output.(string)
+		var description string
+		if ok {
+			description = strMethod
+		} else {
+			mapMethod, ok := output.(map[string]interface{})
+			if ok {
+				description = mapMethod["description"].(string)
+				if mapMethod["severity"] != nil {
+					severityFloat, err := strconv.ParseFloat(mapMethod["severity"].(string), 64)
```
rbren, line 274: oh wow, rego won't give you back the data as a float?

opa/pkg/main.go
```diff
@@ -0,0 +1,324 @@
...
+					}
+					severity = &severityFloat
+				}
+				if mapMethod["title"] != nil {
+					titleString := mapMethod["title"].(string)
+					title = &titleString
+				}
+
+				if mapMethod["remediation"] != nil {
+					remediationString := mapMethod["remediation"].(string)
```
rbren, line 286: should we worry about these not being strings, e.g. if the user did something wrong? Go will `panic` if so

opa/pkg/main.go
baderbuddy: Unfortunately the Rego library returns the string with quotation marks around it. It's super annoying to me. I was trying to find a `Trim` function and couldn't which is why I have this mess. Thanks for this!

opa/pkg/main.go
baderbuddy: I'm going to make it an error. I don't fully understand it, but it really feels like the user passed in the wrong kind of object to this function.

opa/pkg/main.go
baderbuddy: Yes

opa/pkg/main.go
baderbuddy: At this point there's only a couple spots left that the errors are actually floated up. It feels appropriate to me if the API server is temporarily down, or rejects the request or something, to fail the job. Maybe that's not right though....

opa/pkg/main.go
baderbuddy: I did a bunch of refactoring, should remove a ton of indentation.

opa/pkg/main.go
baderbuddy: Rego wants to make life difficult :)

opa/pkg/main.go
rbren: OK yeah I see your point here.

I'd say, if it's something that would break upon the first installation, or upon creating a new policy, it's safe to `panic`. If it's something that could break because of a change in cluster state (e.g. I added a new Widget, which triggered this check for the first time), we should create an action item. Does that sound like a reasonable heuristic?

opa/pkg/main.go
rbren: API server being down definitely deserves an error. I was thinking about RBAC issues, but I defer

opa/pkg/main.go
rbren: Thanks!

PR #220, opened 2020-08-24
Fix rego mistakes in OPA examples

baderbuddy:


opa/examples/label-required.yaml
```diff
@@ -11,8 +11,9 @@ rego: |
           missing := required - provided
           count(missing) > 0
           description := sprintf("Label %v is missing", [missing])
+          severity := 0.1 * count(missing)
           results := {
             description: description,
-            severity: 0.1 * count(missing),
+            severity: severity,
```
rbren, line 17: weird, that's not allowed?

PR #222, opened 2020-08-25
factor out rego package

rbren:
This should make it easier to reuse the validation logic.


baderbuddy:
I think we need to break this up some more. It looks like the only public method in the Rego package does everything through running the rego. We wouldn't want to actually run (because we wouldn't have any input to provide) when validating on the server side. Maybe making `getRegoQuery` public, that feels like it could work.

rbren:
Oh yes, totally meant to do that, thanks

PR #233, opened 2020-08-31
Clean up default repo name, only request origin url when necessary

jordandoig:


ci/pkg/ci/ci.go
```diff
@@ -256,6 +255,9 @@ func SendResults(reports []ReportInfo, resources []Resource, configurationObject
 			originSplit := strings.Split(origin, "//")
 			origin = originSplit[len(originSplit)-1]
 		}
+		// Remove prefix and suffix to get clean $org/$repo structure
+		origin = strings.TrimPrefix("github.com:")
```
baderbuddy, line 259: We probably want to make this more generic, I'd bet that anything that contains a `:` is probably the same thing but it could be a `gitlab.com` or `bitbucket.com` or whatever.

jordandoig: Cool, saw the failing version change test, I'll make this update as well!

PR #237, opened 2020-09-04
Only build changed plugins

rbren:
Build currently takes 20m - this should cut down on that.


.circleci/config.yml
```diff
@@ -29,9 +29,11 @@ commands:
     - run:
         name: Build Plugin Docker Images
         command: |
-          for filename in deploy/*.config; do
-            docker-pull -f $filename
-            docker-build -f $filename
+          for dir in `find ./plugins -type d`; do
+            if ! git diff --exit-code origin/master "$dir/" ; then
+              docker-pull -f $filename
```
baderbuddy, line 34: Looks like a typo here referencing $filename

baderbuddy (approved):
Oh that is clever. Are the e2e tests still going to work?

.circleci/config.yml
rbren: good catch

rbren:
hmm I didn't think about e2e...I think if we grab `version.txt` for the tags instead of $CI_SHA1 it should work

baderbuddy:
If we use the version in version.txt it'll only test master. Because feature branches don't overwrite the version number tags. We could do logic to check if that folder changed and decide from there which tag to use.

rbren:
Ahh good call. Added a helper script to do the change detection, think i've got it but it's hard to tell since every directory changed in this PR

PR #238, opened 2020-09-04
Admission Controller

baderbuddy:


.circleci/config.yml
```diff
@@ -107,6 +107,7 @@ jobs:
       - run: go get -u golang.org/x/lint/golint
       - run: go list ./... | xargs golint -set_exit_status
       - run: cd plugins/opa ; go test ./... ; cd ../..
+      - run: cd plugins/admission ; go test ./... ; cd ../..
```
rbren, line 110: 🎉

plugins/admission/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM golang:1.14.3 AS build-env
+WORKDIR /go/src/github.com/fairwindsops/insights-plugins/
+
+
+ENV GO111MODULE=on
+ENV CGO_ENABLED=0
+ENV GOOS=linux
+ENV GOARCH=amd64
+
+COPY opa opa
```
rbren, line 10: should we move this down below the `copy go.*`? That way if we change the `opa` code the deps are still cached.

plugins/admission/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+WORKDIR /go/src/github.com/fairwindsops/insights-plugins/admission
+
+COPY ./admission/go.* ./
+RUN go mod download
+
+RUN go get -u github.com/gobuffalo/packr/v2/packr2
+
+COPY ./admission/pkg pkg
+COPY ./admission/cmd cmd
+RUN packr2 build /go/pkg/mod/github.com/fairwindsops/polaris*/main.go
```
rbren, line 21: whats with the `*` here?

plugins/admission/cmd/admission/main.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+	var tempConfig models.Configuration
+	err = json.Unmarshal(body, &tempConfig)
+	if err != nil {
+		return err
+	}
+	if tempConfig.Polaris == nil {
+		polarisConfig, err := polarisconfiguration.ParseFile("")
+		tempConfig.Polaris = &polarisConfig
```
rbren, line 66: should probably move this below the error check

plugins/admission/cmd/admission/main.go
```diff
@@ -0,0 +1,132 @@
...
+			return err
+		}
+	}
+	logrus.Infof("Config: %+v", tempConfig)
+	handler.Config = tempConfig
+	return nil
+}
+
+func keepConfigurationRefreshed(ctx context.Context) {
+	ticker := time.NewTicker(time.Minute * 30)
```
rbren, line 77: can we make this configurable?

I feel like `time.Minute` might be a better default - that way when the user changes something server-side they see it reflected almost immediately. Potentially important sec-wise too.

plugins/admission/pkg/admission/admission.go
```diff
@@ -0,0 +1,109 @@
...
+	req.Header.Set("Content-Type", w.FormDataContentType())
+	req.Header.Set("Authorization", "Bearer "+token)
+
+	for _, report := range reports {
+		req.Header.Set("X-Fairwinds-Report-Version-"+report.Report, report.Version)
+	}
+
+	client := &http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
```
rbren, line 66: what's this doing?

plugins/admission/pkg/admission/admission.go
```diff
@@ -0,0 +1,109 @@
...
+var cluster string
+
+func init() {
+	organization = os.Getenv("FAIRWINDS_ORGANIZATION")
+	hostname = os.Getenv("FAIRWINDS_HOSTNAME")
+	cluster = os.Getenv("FAIRWINDS_CLUSTER")
+}
+
+// SendResults sends the results to Insights
+func SendResults(reports []models.ReportInfo, token string) (bool, []string, []string, error) {
```
rbren, line 29: I'm starting to lean towards naming the return values when there are several or it's not obvious, as a way of self-documenting. Would have saved me from a bug or two in the past, but YMMV

plugins/admission/pkg/opa/opa.go
```diff
@@ -0,0 +1,88 @@
...
+					return opa.KubeTarget{
+						APIGroups: []string{splitValues[0]},
+						Kinds:     []string{splitValues[1]},
+					}
+				}).([]opa.KubeTarget),
+			},
+		}
+		found := false
+		for _, target := range instance.Spec.Targets {
+			if apiGroup == target.APIGroups[0] && resourceKind == target.Kinds[0] {
```
rbren, line 44: Should we worry about there being more than one apigroup/kind?

plugins/admission/Dockerfile
baderbuddy: Unfortunately because OPA is one of the dependencies it needs to be there before the go get.

plugins/admission/Dockerfile
baderbuddy: This is definitely a little hacky. But go get appends a name to the directory based on the version. And running packr on just the admission controller doesn't process the dependencies.

plugins/admission/cmd/admission/main.go
baderbuddy: Oh dang! Yeah.

plugins/admission/pkg/admission/admission.go
baderbuddy: Nothing, bad copypasta.

plugins/admission/pkg/admission/admission.go
baderbuddy: Ohh, I actually didn't realize I could do that. I'm totally changing this.

plugins/admission/pkg/opa/opa.go
baderbuddy: Ideally, in this case though those arrays were just created by a request to Insights that returns an array of strings where it's `group/kind` and then splitting them. So in this particular case until/if we update Insights to store an array of groups and an array of kinds this is all we'll get back.

plugins/admission/cmd/admission/main.go
baderbuddy: 30 minutes is probably a bit much :) for now making it configurable seems like a good idea. I think ultimately keeping track of when something in the config last changed and returning a short response if it's the same version would be the way to go. I worry that with the Rego being in there this config could easily get pretty long.

rbren (approved):

PR #249, opened 2020-09-11
Add container level information to CI/CD

baderbuddy:
This saves a list of containers found in a pod in the Workloads report. It also sends owner information for images that aren't in a manifest and container level information for images that are in a manifest.


plugins/ci/pkg/ci/ci.go
```diff
@@ -109,19 +118,22 @@ func GetImagesFromManifest(configFolder string) ([]models.Image, []Resource, err
...
 	images := getImages(podSpec.(map[string]interface{}))
-	return funk.Map(images, func(s string) models.Image {
+	return funk.Map(images, func(c Container) models.Image {
 		return models.Image{
-			Name:  s,
-			Owner: owner,
+			Name: c.Image,
+			Owner: models.Resource{
+				Kind: owner.Kind,
+				Name: fmt.Sprintf("%s/%s", owner.Name, c.Name),
```
rbren, line 133: Do we need to put the container name in the `Owner`? I feel like `Owner` refs should reliably map to k8s resources...

baderbuddy: This is keeping it consistent with how the Trivy scan works right now. It uses the "Controller/container" name that gets saves in Insights.

rbren: OK. We have to get rid of this pattern before it gets too pervasive 😓

rbren (approved):

PR #283, opened 2020-10-02
 adjust logs for admission controller

rbren:


makoscafee (approved):
LGTM

baderbuddy (approved):
This looks pretty good. Do we actually set the logging level anywhere? I feel like we should probably default to warning but allow people to go to info.

rbren:
Hmm good point, we should definitely allow setting it.

I tend to stick with INFO-ish as default, because it makes it easier for us to help customers debug when something goes wrong, but could be convinced that WARN is better.

PR #286, opened 2020-10-16
 update trivy to latest

rbren:


baderbuddy (approved):
Ohh good call on the version numbers. Do we want to update Trivy in CI at the same time?

makoscafee (approved):

rbren:
Good call! updated CI

PR #287, opened 2020-11-09
[opa] Adding additional policies

baderbuddy:
Restructured the examples to fit into the directory format for the forthcoming sync utility.

Added a few new policies.

Added the ability to block certain namespaces (such as kube-system) from the results of a policy.

Added the clusters to the CheckSetting type to make it easier to use in the future.


plugins/opa/examples/fresh-deployment/policy.yaml
```diff
@@ -0,0 +1,24 @@
...
+        }
+        latestReplicaDateStale(replicaSets, elem) {
+          rs := replicaSets[_]
+          owner := rs.metadata.ownerReferences[_]
+          owner.uid == elem.metadata.uid
+          rs.status.replicas > 0
+          comparisonDate := time.add_date(time.now_ns(), 0, 0, 0 - input.parameters.days)
+          time.parse_rfc3339_ns(rs.metadata.creationTimestamp) < comparisonDate
+        }
+        hparequired[description] {
```
rbren, line 20: `hpareqiured` should probably be `deploymentIsStale` or similar

plugins/opa/examples/fresh-deployment/policy.yaml
```diff
@@ -0,0 +1,24 @@
...
+          not blockedNamespace(input)
```
rbren, line 21: I wonder if we should be putting this logic into the configuration rather than the rego - seems like it'd be pretty common

rbren (approved):
Looking forward to being able to test these with the CLI eventually!

PR #290, opened 2020-11-20
[prometheus] Add a new plugin for collecting Prometheus Metrics.

baderbuddy:
Adds a new plugin to collect Prometheus Metrics.

Also fixes the flakey results on the Trivy e2e tests (hopefully)


plugins/prometheus/pkg/data/data.go
```diff
@@ -0,0 +1,179 @@
...
+		request.memoryLimit = memVal.Values[0].Value
+		combinedRequests[key] = request
+	}
+	requestArray := make([]CombinedRequest, 0, len(combinedRequests))
+	workloads, err := controller.GetAllTopControllers(ctx, dynamicClient, restMapper, "")
+	if err != nil {
+		return nil, err
+	}
+	for _, val := range combinedRequests {
+		val.ControllerName, val.ControllerKind = getController(workloads, val.PodName, val.ControllerNamespace)
```
rbren, line 163: It would probably be more efficient to create a map[podName]controller, rather than iterating through `workloads` every time. That array could get huge in big clusters.

plugins/prometheus/pkg/data/types.go
```diff
@@ -0,0 +1,50 @@
...
+
+// CombinedRequest is the cpu/memory and requests for a given pod/container
+type CombinedRequest struct {
+	Owner
+	cpu           []model.SamplePair
+	memory        []model.SamplePair
+	memoryRequest model.SampleValue
+	cpuRequest    model.SampleValue
+	memoryLimit   model.SampleValue
+	cpuLimit      model.SampleValue
```
rbren, line 49: is it possible to observe changes in limits/requests over time? That'd be a huge increase in data storage, but could make things more accurate...

rbren (approved):

plugins/prometheus/pkg/data/types.go
baderbuddy: Yep, we're going to be sending up requests/limits for each so we'll be able to visualize that.

rbren: So why are these a single value, versus slices above?

baderbuddy: Oh because an individual pod's request/limits shouldn't ever change. So we'll track the request/limits changes of the controller over time by capturing the request/limits of these individual pods. There might be some edge cases with static pods or naked pods where a new pod with the same name is created. But that should only really mess with data like right at the margins of the 10 minute windows we're running this job on. https://github.com/FairwindsOps/insights-plugins/blob/fe07030c930d7249daecce86651e9450088b875d/plugins/prometheus/pkg/data/types.go#L32 is the actual type getting saved off. CombinedRequest is just a helper class to merge the data to make it easier to process (probably not even that useful anymore)

rbren: Ohh that makes so much more sense. Thanks!

PR #310, opened 2020-12-10
[ci] more logging/output

rbren:
## Description
Adds some more logs to stdout

## Checklist


baderbuddy (approved):
This looks good to me. I'm curious why you went with a minor version bump instead of patch version bump though.

rbren:
🤷 junit output change? I could be convinced otherwise

PR #312, opened 2020-12-21
Fix CI Issues

baderbuddy:
## Description

Dedupes Trivy scans to make CI run faster.

Fixes OPA to not break in CI if multiple manifests modify the same object.

Updates the kubectl version to a stable release

## Checklist


plugins/ci/cmd/insights-ci/main.go
```diff
@@ -212,18 +212,23 @@ func getTrivyReport(images []trivymodels.Image, configurationObject models.Confi
...
 		return trivyReport, err
 	}
-
+	refLookup := map[string]string{}
 	// Download missing images
 	for idx, currentImage := range images {
 		if currentImage.PullRef != "" {
 			continue
 		}
+		if ref, ok := refLookup[currentImage.Name]; ok {
```
rbren, line 221: guessing `Name` has the sha, not just the tagname?

plugins/opa/pkg/kube/files.go
```diff
@@ -74,7 +75,10 @@ func SetFileClient(objects []map[string]interface{}) *Client {
 		}
 		_, err = dynamic.Resource(gvr).Namespace(namespace).Create(context.Background(), &unstructure, metav1.CreateOptions{})
 		if err != nil {
-			panic(err)
+			statusError, ok := err.(*kubeerrors.StatusError)
+			if !ok || statusError.ErrStatus.Reason != metav1.StatusReasonAlreadyExists {
```
rbren, line 79: I think we need to do a better job of handling this case. As-is, the second file won't be scanned, and the user won't get any feedback to that effect. I'd rather `panic` and not support this case than let a failure slip through.

Assuming it's not a trivial fix, this is probably worth breaking out into its own PR

plugins/ci/cmd/insights-ci/main.go
baderbuddy: It's the tagname (or whatever is specified in the manifest). But that's all we have in the CI/CD context. It's exactly the same thing we use for the skopeo copy below.

plugins/opa/pkg/kube/files.go
baderbuddy: The second file will be scanned. This only affects the in-memory kubernetes cluster that an OPA policy could potentially use (but I can't see it being a common thing in the CI context anyways). Assuming most of those checks are just checking for the existence of something this actually feels like a safe way to handle it to me.

baderbuddy: a panic here really doesn't feel appropriate. But we could easily log it. Or something like that. I'm okay with breaking this into a different PR. I needed it in here to test out with the Charts repo though

rbren: Ohh I understand now. OK, yeah, I think this is good then.

plugins/ci/cmd/insights-ci/main.go
rbren: 👍

rbren (approved):

PR #313, opened 2021-01-04
replace helm release name in the configfolder path with helm path

makoscafee:
## Description

Replaces the helm release name in the folder name with helm path.

## Checklist


plugins/ci/pkg/ci/ci.go
```diff
@@ -459,11 +459,15 @@ func ProcessHelmTemplates(configurationObject models.Configuration, configFolder
 		if err != nil {
 			return err
 		}
+		filefolder := configFolder
+		if strings.HasPrefix(filefolder, helmObject.Name) {
+			filefolder = strings.Replace(configFolder, helmObject.Name, helmObject.Path, 1)
```
baderbuddy, line 444: This will just template out the charts to a different directory (I think it might template them overtop where the chart is). The checking for prefix needs to happen in GetAllResources that way we can keep the `helmObject.Name` so the same chart could be processed multiple times for different values files.

makoscafee: would this place here be the right place? 
https://github.com/FairwindsOps/insights-plugins/blob/main/plugins/ci/pkg/ci/ci.go#L61

baderbuddy: Oh yeah that looks perfect!

plugins/ci/pkg/ci/ci.go
```diff
@@ -50,15 +50,21 @@ func GetResultsFromCommand(command string, args ...string) (string, error) {
...
 	images := make([]trivymodels.Image, 0)
 	resources := make([]models.Resource, 0)
 	err := filepath.Walk(configFolder, func(path string, info os.FileInfo, err error) error {
 		if !strings.HasSuffix(info.Name(), ".yaml") {
 			return nil
 		}
-		relativePath, err := filepath.Rel(configFolder, path)
+		filefolder := configFolder
+		for _, helmObject := range configurationObject.Manifests.Helm {
+			if strings.HasPrefix(filefolder, helmObject.Name) {
```
rbren, line 63: I think this would be a bit safer if we did `strings.HasPrefix(filefolder, helmObject.Name + "/")`

rbren: If we hit this block, can we save `helmObject.Name` as `resource.HelmName` (at lines 106 and 124 below)

rbren:
@makoscafee I have no idea why the test is failing. It does seem to be specific to your PR though

rbren:
@baderbuddy any ideas?

baderbuddy:
It definitely looks like the workloads report's json schema. I can't figure out for the life of me why the workloads changed though. If we didn't use the rok8s scripts job for the e2e tests then we could store the artifacts of the JSON responses so we could easily compare them ourselves against the schema and fix any issues. The rok8s script job has the ability to use a JUNIT output, we could possibly structure the results into a JUNIT file. I don't know for sure how to handle that yet though.

rbren (approved):

rbren:
Tests passing! Sorry for the breakage

PR #316, opened 2021-02-03
 add instances for each policy

rbren:
## Description

Add instances for all policies

## Checklist


plugins/opa/examples/hpa-required/workloads.yaml
```diff
@@ -0,0 +1,7 @@
+targets:
+- apiGroups: ["apps"]
+  kinds: ["DaemonSet", "Deployment", "StatefulSet"]
```
baderbuddy, line 3: Let's bring this down to just deployments. An hpa on the others doesn't feel like it makes sense

baderbuddy (approved):
One small change and that's it

plugins/opa/examples/hpa-required/workloads.yaml
rbren: Hah very good point

PR #334, opened 2021-03-01
Change  default option TempFolder to tmp

makoscafee:
## Checklist


plugins/ci/pkg/models/models.go
```diff
@@ -116,7 +116,7 @@ func maybeAddSlash(input string) string {
 // SetDefaults sets configurationd defaults
 func (c *Configuration) SetDefaults() {
 	if c.Options.TempFolder == "" {
-		c.Options.TempFolder = "./_insightsTemp/"
+		c.Options.TempFolder = "/tmp/"
```
baderbuddy, line 119: I feel like it might be safer to have `/tmp/<something>` like maybe still keep the _insightsTemp name, but just inside `/tmp/` instead of `.`

baderbuddy (approved):
LGTM, but I recommended a change in the folder name.

PR #357, opened 2021-03-09
Disable reports on CI/CD

baderbuddy:
## Description

Adds new configuration options to disable one report at a time.

## Checklist


rbren:
I think Andy's issue still stands where it will try to pull the image, no?

baderbuddy:
> I think Andy's issue still stands where it will try to pull the image, no?

@rbren I don't think so. It looks like images aren't actually pulled until the getTrivyReport function, so if he disabled Trivy then it shouldn't even attempt to pull.

rbren:
🤔 don't we pull them in the insights-ci script?

baderbuddy:
We only pull/save the ones that are specified as just "images" not ones from the manifests

baderbuddy:
I'm pretty sure

rbren:
nevermind, reread the message, I think you're right

PR #360, opened 2021-03-11
Add script for downloading config file

baderbuddy:
## Description

Adds a new download config script to the uploader image. This will be used for downloading config files from Insights to be used by the plugins.

## Checklist


makoscafee (approved):
LGTM

plugins/uploader/download.sh
```diff
@@ -0,0 +1,78 @@
...
+then
+    set +x
+fi
+usage()
+{
+cat << EOF
+usage: FAIRWINDS_TOKEN=secret $0 \
+    --organzation acme-co \
+    --cluster staging \
+    --file foo.json \
```
makoscafee, line 15: I know this is just an example... but I think the file type is supposed to be `.yaml`?

plugins/uploader/download.sh
```diff
@@ -0,0 +1,78 @@
...
+file=''
+datatype=''
+timeout='60'
+host='https://insights.fairwinds.com'
+
+while [ ! $# -eq 0 ]; do
+    flag=${1##-}
+    flag=${flag##-}
+    value=${2}
+    case "$flag" in
```
makoscafee, line 34: Do you think there is a need to check if `FAIRWINDS_TOKEN` was supplied?

makoscafee:
I think I have more questions. will `download.sh` used  together with `upload.sh`? if not would it make sense for it to have a different plugin folder or probably a `README` section update?

baderbuddy:
@makoscafee I wanted to avoid doing a separate plugin because it would add extra configuration in the chart and also require creating another Dockerfile/Quay repo/etc and that seemed like overkill for the same set of dependencies just slightly different logic.

makoscafee:
Oh ok, Makes sense

PR #371, opened 2021-03-16
Add external-probes policy

ivanfetch:
This policy matches pod specifications that specify an external host in a livenessProbe or readinessProbe, which may be undesirable RE: [Pod probes lead to blind SSRF from the node  #99425](https://github.com/kubernetes/kubernetes/issues/99425).

The `checkPod` rule avoids checking pods with a
`metadata.ownerReference`, to avoid duplicate action-items for pods that belong to
other pod-controllers also checked by this same policy.

This was tested against:

* A deployment with a single container, whos readinessProbe targets an external host.
* A deployment whos second container has a readinessProbe targeting an
  external host.
* A single pod whos readinessProbe targets an external host.
* A cronJob with a readinessProbe targeting an external host.

## Checklist


baderbuddy (approved):
Looks great to me!

PR #378, opened 2021-03-26
Fix download script to allow 404

baderbuddy:
## Description

Fix the download script, if there's a 404 right now it will fail.

## Checklist


rbren (approved):

plugins/uploader/download.sh
```diff
@@ -63,13 +63,14 @@ fi
...
 set +x
+status=0
 results=$(curl -X POST $url \
   -L \
   -H "Authorization: Bearer ${FAIRWINDS_TOKEN//[$'\t\r\n']}" \
   -H "Accept: application/x-yaml" \
   -O $file \
-  --fail 2>&1)
-if [ $? -ne 0 ]
+  --fail 2>&1) || status=$?
```
rbren, line 72: why not just remove `--fail`?

baderbuddy: So that I can still fail for a 500 or 400 or other status code. It feels like that's still the appropriate choice. I could change course on that though.

plugins/uploader/download.sh
```diff
@@ -63,13 +63,14 @@ fi
...
+if [ $status -ne 0 ]
 then
     echo $results
     # If the response isn't a 404 this job will fail
```
rbren, line 76: should the grep below be `grep -v` (inverted)? Not sure how --quiet works

plugins/uploader/download.sh
rbren: 👍 makes sense

plugins/uploader/download.sh
baderbuddy: It feels right to me, but maybe it's just too early. That should fail if it finds something other than 404. That way a 200 would just skip that if statement and output a file. a 404 would exit successfully and anything other than a 404 would throw an error and fail the pod.

baderbuddy:
There's a bug in the workloads e2e test but I'm going to go ahead and merge this because that's not related. I'm going to see if I can get a change into rok8s scripts to make those things easier to debug.

plugins/uploader/download.sh
rbren: ohh I understand now. 👍

PR #449, opened 2021-07-19
Update some go dependencies

rbren:
## Description

Some deps (e.g. controller-runtime) are having a hard time, but this is a first step.

## Checklist


plugins/workloads/go.mod
```diff
@@ -1,12 +1,11 @@
-module github.com/fairwindsops/insights-plugins/workload
+module github.com/fairwindsops/insights-plugins/workloads
```
makoscafee, line 1: you probably have to change folder name to `workload` here

rbren: I think it's fine - we're not using this as a go library anywhere. LMK if you think otherwise

.circleci/config.yml
```diff
@@ -118,12 +118,19 @@ jobs:
...
       - checkout
       - set_environment_variables
       - run: go get -u golang.org/x/lint/golint
-      - run: go list ./... | xargs golint -set_exit_status
+      - run: |
+          for plugin in "${GO_PKGS[@]}"; do
+            echo "linting $plugin"
+            cd plugins/$plugin
+            # TODO: fix lint errors
+            # go list ./... | xargs golint -set_exit_status
```
baderbuddy, line 131: Was the listing not actually happening before?

baderbuddy (approved):
LGTM

.circleci/config.yml
rbren: Yeah I think it was just empty, probably due to moving everything into the `plugins` directory. Go 1.16 exits non-zero.

PR #450, opened 2021-07-20
Bump version of Rok8s scripts and store artifacts

baderbuddy:
## Description

Stores output files from e2e tests. If there's a failure then you can grab the JSON file that failed and bump it against the schema manually to try to figure out what's wrong with the schema file.

## Checklist


rbren (approved):
👍 thanks! I was stuck 😄

PR #483, opened 2021-09-01
[ci] add SHAs to trivy images

makoscafee:
## Checklist


plugins/ci/pkg/ci/ci.go
```diff
@@ -220,30 +220,70 @@ func GetRepoTags(path string) ([]string, error) {
...
+			return "", err
+		}
+		var jsonBody interface{}
+		err = json.Unmarshal(bytes, &jsonBody)
+		if err != nil {
+			return "", err
+		}
+		var sha string
+		configMap := jsonBody.(map[string]interface{})["config"].(map[string]interface{})
+		sha = configMap["Image"].(string)
```
rbren, line 283: is there any chance this field won't exist? If so we should handle the `nil` case.

plugins/ci/cmd/insights-ci/main.go
```diff
@@ -217,7 +221,9 @@ func getTrivyReport(images []trivymodels.Image, configurationObject models.Confi
 		if !matchedImage && len(repoTags) > 0 {
 			imageRepo := repoTags[0]
 			imageRepo = strings.Split(imageRepo, ":")[0]
+			ID := fmt.Sprintf("%s@%s", imageRepo, sha)
```
rbren, line 224: I think this needs to be `@sha256:`

rbren:
🤔 I think we do need my (old) PR for this one. This will only set SHA for images that were explicitly in the fairwinds-insights.yaml, not images that were found in the Helm/YAML

plugins/ci/cmd/insights-ci/main.go
makoscafee: mmh aparently `config.Image` has this format `sha256:xxxxxx`
```
 "ArgsEscaped": true,
        "Image": "sha256:2bf39c9ea98d819cb507e965ebebdc87123b348d5127282c24782d43be52fc00",
        "Volumes": null,
        "WorkingDir": "",
```

rbren: aha! good to know, thanks

rbren (approved):

PR #508, opened 2021-09-29
Added Limit and Request values

makoscafee:
## Checklist


makoscafee:
sample report -> https://gist.github.com/makoscafee/eab1bea8297afbaa067af1c511f6d866

plugins/prometheus/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+## 0.4.6
+* Added Request and Limit values
+* Modify resource id format to cater for new prometheus and kubernetes versions(v1.21.1)
```
rbren, line 4: 🎉 good find!

Do we need to put a switch in the code based on the Kubernetes version?

makoscafee: i think not ... old kubernetes pods id had this format `/kubepods/xxxxx` but new version starts with whatever manages that pod `{docker/kubelet}/kubepods` so `.*/kubepods.*` will cover both of them. But let me try for `1.16` to `1.20` and see what i get.

rbren: OK, I was thinking more about the change from

```
ube_pod_container_resource_requests_memory_bytes{container!="POD", container!=""}
```
to
```
kube_pod_container_resource_requests{container!="POD", container!="", unit="byte", resource="memory"}
```
but maybe that was never working? It's not working for us in 1.19 at least

makoscafee: so tested this in Kubernetes `v1.16.15`, `v1.17.17`, `v1.19.11` and `v1.21.1` and prometheus version `13.2.0`, `13.2.1`,(the we are using now) and  `14.8.0`(latest) and all of them do not have `kube_pod_container_resource_requests_memory_bytes` but `kube_pod_container_resource_requests` ... not sure which version of kubernetes and/or prometheus displayed data using `kube_pod_container_resource_requests_memory_bytes `

makoscafee: Good news is tested with p8s version `10.6.0`  and `11.16.9` so they seem to have both `kube_pod_container_resource_requests ` and  `kube_pod_container_resource_requests_memory_bytes`.. the latter support seemed to be discounted  in either`12` or `13`

rbren: Patch release it is!

rbren (approved):

PR #518, opened 2021-10-19
Added Falco agent

makoscafee:
## Description
This plugin has two endpoints
- `/data` a webhook endpoint that receives data from falco
- `/output` an endpoint that is used to retrieve aggregated data for the past 24 hrs.

## Checklist


plugins/falco-agent/pkg/routes.go
```diff
@@ -0,0 +1,110 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	payload, err := data.Aggregate24hrsData(outputfolder)
+	if err != nil {
+		logrus.Errorf("Error while aggregating data: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	requestArray := make([]data.FalcoOutput, 0, len(payload))
+
+	workloads, err := controller.GetAllTopControllers(ctx, dynamicClient, restMapper, "")
```
rbren, line 53: Per offline discussion, I think this will be more efficient if we just get one controller at a time for the pods we're interested in

rbren: Actually - it would probably be better to gather this data when the events come _in_, and to save it to disk.

That way if a controller got deleted, we'll still have the data.

rbren (approved):

PR #532, opened 2021-11-16
Use controller util to get all workloads in the cluster

makoscafee:
## Checklist


vitorvezani (approved):
LGTM! Very generic, clean and concise 🔥

ivanfetch-fw (approved):
Wow - that's quite an improvement that replaces a lot of internal plugin code to track / map / determine individual "top controllers"  by our `GetAllTopControllers` function!

plugins/workloads/pkg/info.go
```diff
@@ -123,250 +127,32 @@ func CreateResourceProviderFromAPI(ctx context.Context, kube kubernetes.Interfac
...
-			ownerController := controllerMap[ownerUID]
-			if ownerController != nil {
-				if item.Status.StartTime != nil && item.Status.CompletionTime != nil {
-					cronChildren[ownerUID] = append(cronChildren[ownerUID], jobMetadata{
-						startTime: item.Status.StartTime.Time,
-						endTime:   item.Status.CompletionTime.Time,
-					})
-				}
-				continue
+			for _, ctn := range pd.Spec.Containers {
```
rbren, line 148: I think we'll end up getting the same containers for each pod - we really just want to know what the container names are. 

You could use [`GetPodSpec`](https://github.com/FairwindsOps/controller-utils/blob/e9a28b6e43fcb89e2d304d5004566c8c5c383181/pkg/controller/pod.go#L20) to get the container names.

rbren (changes requested):
Thanks Barnie! Love to see all that deleted code.

I think we need to change the container logic a bit though

rbren (approved):

PR #551, opened 2021-12-07
FWI-1017 - Initial updating of memory limits

ivanfetch-fw:
## Description
Update the right-sizer controller to:
* Handle a `/healthz` HTTP endpoint for readiness/liveness probes, returning the number of items in the report to help ensure there isn't a mutex deadlock on the report data structure.
* Modify memory limits of pod-controllers in response to their pods having been OOM-killed, including command-line configurability for enabling limit-updating, specifying increment and maximum multipliers for updates.
* Add a global namespaces filtering option (applies to action-items and updating memory limits).

Additional details are in [the ticket](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1017). See also [the accompanying Helm chart PR](https://github.com/FairwindsOps/insights-plugins/pull/551).

## Command-line Help Output

Here are the available command-line options, via `right-sizer -h`:

```
Application Options:
  -v, --verbose=                        Show verbose debug information [$VERBOSE]
  -S, --state-configmap-namespace=      The Kubernetes namespace for the ConfigMap resource used to persist report state inbetween controller restarts. [$RIGHTSIZER_STATE_CONFIGMAP_NAMESPACE]
  -s, --state-configmap-name=           The name of the ConfigMap resource used to persist report state inbetween controller restarts. [$RIGHTSIZER_STATE_CONFIGMAP_NAME]
  -N, --namespace=                      Only respond to OOM-killed containers in these Kubernetes namespaces. By default, all namespaces are allowed. This applies both to alerting on OOM-kills and
                                        modifying memory limits if --update-memory-limits is enabled. Use this option multiple times to specify multiple namespaces. IF setting namespaces via the
                                        environment variable, separate namespaces by a comma. [$RIGHTSIZER_NAMESPACES]
  -w, --reset-ooms-window=              The amount of time after which an item will be removed from the report, if no OOM-kills have been seen. Specify this as a time duration, such as 24H.
                                        [$RIGHTSIZER_RESET_OOMS_WINDOW]
  -m, --update-memory-limits            Update the memory limits of resources, such as deployments, that own pods that have been OOM-killed. Updating stops once the threshold is reached defined by the
                                        --update-memory-limits-max option. To specify which namespaces are allowed to be updated, see the --namespaces option. [$RIGHTSIZER_UPDATE_MEMORY_LIMITS]
  -n, --update-memory-limits-min-ooms=  The number of OOM-kills required before the memory limits will be updated. This option is only used if --update-memory-limits is enabled.
                                        [$RIGHTSIZER_UPDATE_MEMORY_LIMITS_MIN_OOMS]
  -u, --update-memory-limits-increment= The multiplier used to increase memory limits, in response to an OOM-kill. This value is multiplied by the limits of the OOM-killed container. This option is
                                        only used if --update-memory-limits is enabled. IF this option is specified, the update-memory-limits-max must also be specified.
                                        [$RIGHTSIZER_UPDATE_MEMORY_LIMITS_INCREMENT]
  -U, --update-memory-limits-max=       The multiplier used to calculate the maximum value to update memory limits. This value is multiplied by the starting memory limits of the first OOM-killed
                                        container seen by this controller. This option is only used if --update-memory-limits is enabled. If this option is specified, the
                                        --update-memory-limits-increment must also be specified. [$RIGHTSIZER_UPDATE_MEMORY_LIMITS_MAX]


```

## Checklist

[YouTrack FWI-1017](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1017)


plugins/right-sizer/cmd/main.go
```diff
@@ -1,25 +1,55 @@
...
 	"github.com/fairwindsops/insights-plugins/right-sizer/src/controller"
+	"github.com/fairwindsops/insights-plugins/right-sizer/src/report"
 	"github.com/fairwindsops/insights-plugins/right-sizer/src/util"
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 var opts struct {
-	Verbose int `env:"VERBOSE" short:"v" long:"verbose" description:"Show verbose debug information"`
+	Verbose                       int           `env:"VERBOSE" short:"v" long:"verbose" description:"Show verbose debug information"`
```
rbren, line 17: interesting that it puts all this stuff in annotations

plugins/right-sizer/cmd/main.go
```diff
@@ -1,25 +1,55 @@
...
+	StateConfigMapNamespace       string        `env:"RIGHTSIZER_STATE_CONFIGMAP_NAMESPACE" short:"S" long:"state-configmap-namespace" description:"The Kubernetes namespace for the ConfigMap resource used to persist report state inbetween controller restarts."`
```
rbren, line 18: Does this default to its own namespace?

plugins/right-sizer/cmd/main.go
```diff
@@ -1,25 +1,55 @@
...
+
 	stopChan := make(chan struct{})
-	eventGenerator := controller.NewController(stopChan)
-	util.InstallSignalHandler(eventGenerator.Stop)
+	kubeClientResources := util.CreateKubeClientResources() // Create client, dynamic client, RESTMapper...
+	reportBuilder := report.NewRightSizerReportBuilder(kubeClientResources,
+		report.WithStateConfigMapNamespace(opts.StateConfigMapNamespace),
+		report.WithStateConfigMapName(opts.StateConfigMapName),
+		report.WithTooOldAge(opts.ResetOOMsWindow),
+	)
```
rbren, line 42: love this pattern

plugins/right-sizer/src/util/helpers.go
```diff
@@ -0,0 +1,11 @@
...
+
+// Contains returns true if the string is present in the slice.
+func Contains(list []string, wantedItem string) bool {
+	for _, item := range list {
+		if item == wantedItem {
+			return true
+		}
+	}
+	return false
+}
```
rbren, line 11: We like to use [go-funk](https://github.com/thoas/go-funk) for stuff like this

plugins/right-sizer/src/util/kubernetes.go
```diff
@@ -68,3 +83,194 @@ func Clientset() KubeClientResources {
...
+		Object: objectAsMap,
+	}
+
+	topController, err := fwControllerUtils.GetTopController(context.TODO(), kubeClientResources.DynamicClient, kubeClientResources.RESTMapper, unstructuredPod, nil)
+	if err != nil {
+		return nil, err
+	}
+	glog.V(2).Infof("found controller kind %q named %q for pod %s/%s", topController.GetKind(), topController.GetName(), pod.Namespace, pod.Name)
+	return &topController, nil
+}
```
rbren, line 251: this stuff will end up in controller-utils right?

plugins/right-sizer/version.txt
```diff
@@ -1 +1 @@
-0.1.0
+0.1.1
```
rbren, line 1: I would suggest `0.2`, unless you want to make sure this change gets shipped out to our cluster without needing the team to update the chart.

rbren (approved):

plugins/right-sizer/src/util/helpers.go
ivanfetch-fw: I've made that change and removed my own `Contains` function.

plugins/right-sizer/src/util/kubernetes.go
ivanfetch-fw: Yes that's the eventual plan. I haven't done it so far because the functions I'm creating for right-sizer keep changing enough IMO.

PR #558, opened 2021-12-16
Adding opa policy to block rollout max unavailability of > than 25%

rwsweeney:
## Description

This policy checks to see if the maxUnavailable value for a rollingdeployment is greater than 25%, and if it is it creates an action item.

## Checklist

* [X] I have updated the CHANGELOG of any plugins updated
* [X] I have updated the version file of any plugins updated
* [X] I have added tests wherever appropriate
* [X] I have labeled this PR with a label per plugin updated


plugins/opa/examples/rollout-strategy/policy.rego
```diff
@@ -0,0 +1,30 @@
...
+  not blockedNamespace(input)
+  
+  # These lines check to see if the maxUnavailable field is a percentage, and if so to strip it and converts it to a number for evaluation.
+  maxUnavailable := input.spec.strategy.rollingUpdate.maxUnavailable
+  contains(maxUnavailable, "%")
+  maxUnavailble = replace(maxUnavailable, "%", "")
+  newMaxUnavailable = to_number(maxUnavailble)
+  
+  # If newMaxUnavailable is greater than 25 then create an action item.
+  newMaxUnavailable = 25
```
rbren, line 21: would suggest making this `==` per [these docs](https://www.openpolicyagent.org/docs/latest/policy-language/#equality-assignment-comparison-and-unification)

Actually I'm not sure if `=` works here? I don't really understand how `=` works in rego.

rbren: Oh actually - reading your comment - I think this should be `newMaxUnavailable > 25`

rwsweeney: Ooh, good catch on newMaxUnavailable > 25! Not sure how that ended up as an =. That should now be fixed.

The single equals sign works in my test cluster, and when I add == for line 17 and 18 I get the following errors:
fairwinds:17: rego_unsafe_var_error: var maxUnavailble is unsafe
fairwinds:18: rego_unsafe_var_error: var newMaxUnavailable is unsafe

plugins/opa/version.txt
```diff
@@ -1 +1 @@
-0.3.9
+0.3.91
```
rbren, line 1: I think this should be `0.3.10`

plugins/opa/examples/rollout-strategy/helm-deployed.yaml
```diff
@@ -0,0 +1,8 @@
+parameters:
```
rbren, line 1: I think this file should be called `deployments.yaml` or similar

rbren (commented):
Thanks Rachel!

rbren (approved):

PR #566, opened 2021-12-21
FWI-1093 - Trivy 0.12.0 does not bring log4j vulns - update to 0.21.3

jdesouza:
## Checklist

[YouTrack FWI-1093](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1093)


rbren (commented):
Thanks James! Let's make sure Barnie agrees with our thinking here tomorrow but I think this will do the trick!

plugins/trivy/pkg/image/scan.go
```diff
@@ -115,7 +115,7 @@ func ScanImageFile(imagePath string, imageID string, tempDir string) ([]models.V
 		os.Remove(reportFile)
 	}()
 
-	report := []models.VulnerabilityList{}
+	report := models.TrivyResults{}
```
makoscafee, line 118: mmh why did you need to wrap this to  new struct?

plugins/trivy/pkg/models/image.go
```diff
@@ -30,6 +30,10 @@ type ImageReport struct {
 	Report         []VulnerabilityList
 }
 
+type TrivyResults struct {
```
makoscafee, line 33: why do we need this?

jdesouza: We update trivy version because old version is not report with all vulns, but trivy report updated their json schema and added a new Results tag.

plugins/trivy/pkg/image/scan.go
jdesouza: Added a Results tag for new trivy report schema version

makoscafee (approved):

PR #574, opened 2022-01-04
FWI-1166 - Update right-sizer E2E test to update memory limits

ivanfetch-fw:
## Description
This updates the right-sizer E2E test to trigger a second OOM-kill, causing the controller to update memory limits. The test verifies the report item reflects the number of OOMs and the updated `endingMemory` value.
* As the controller reflects the post-patch; in-cluster memory value in the report item, the test is validating that the workload deployment was patched.

The test includes many `Kubectl wait` commands to help verify in-cluster resources are healthy, as the CI kind cluster can be less healthy; predictable, running all the Insights jobs.

I've run this test 5 times successfully without flakes, before opening this PR for review.

## Checklist

[Internal Ticket FWI-1166](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1166)


rbren (approved):
Looks great, thanks Ivan!

PR #575, opened 2022-01-05
FWI-1174 - Add update-memory-namespace option to right-sizer controller

ivanfetch-fw:
RE: plugin version, merge this PR, **after** [PR 574](https://github.com/FairwindsOps/insights-plugins/pull/574) is merged.

## Description
Limit the Kube namespaces where memory limits will be updated, via a `--update-namespace` option.

## Checklist

[Internal Ticket FWI-1174](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1174)


plugins/right-sizer/src/controller/controller.go
```diff
@@ -162,6 +171,10 @@ func NewController(stop chan struct{}, kubeClientResources util.KubeClientResour
 func (c *Controller) Run() error {
 	glog.Infof("Controller configuration is: %+v", c.config)
 	glog.Infof("Report configuration is: %s", c.reportBuilder.GetConfigAsString())
+	dissimilarNamespaces := funk.LeftJoinString(c.config.allowedUpdateNamespaces, c.config.allowedNamespaces)
+	if c.config.updateMemoryLimits && len(dissimilarNamespaces) > 0 {
+		glog.Errorf("NOTE: memory limits will not be updated in these Kubernetes namespaces: %v Although these namespaces are in the allowedUpdateNamespaces list, they are missing from the global allowedNamespaces one.", dissimilarNamespaces)
+	}
```
rbren, line 177: I'm not sure I agree with this check - why not just do everything in `allowedUpdateNamespaces`?

Also - don't we have to worry about the case where `allowedNamespaces` is empty (meaning all namespaces)?

plugins/right-sizer/src/controller/controller.go
```diff
@@ -284,34 +297,38 @@ func (c *Controller) processOOMKilledPod(pod *core.Pod) {
...
-						glog.Errorf("error patching %s memory limits: %v", reportItem, err)
-					} else {
-						// The post-patch memory limits may be different than the limits in our patch.
-						// I.E. We patch with 15655155794400u, post-patch shows 15655155795m
-						// Update  report item with the actual; post-patch memory.
-						patchedContainer, _, _, foundPatchedContainer, err := util.FindContainerInUnstructured(patchedResource, reportItem.ResourceContainer)
-						if !foundPatchedContainer || err != nil {
-							glog.Errorf("unable to find the container %s in patched report item %s, and will have to set report memory limits to the value sent in the patch: %s - potential error = %v", reportItem.ResourceContainer, reportItem, newContainerMemoryLimits, err)
-							reportItem.EndingMemory = newContainerMemoryLimits
+			if len(c.config.allowedUpdateNamespaces) > 0 && funk.ContainsString(c.config.allowedUpdateNamespaces, pod.Namespace) {
```
rbren, line 300: I'd suggest trying to break up this function a bit - it's pretty long and deeply nested.

plugins/right-sizer/src/controller/controller.go
```diff
@@ -67,6 +68,14 @@ func WithAllowedNamespaces(value []string) controllerOption {
 	}
 }
 
+// WithAllowedUpdateNamespaces sets the corresponding field in a controllerConfig type.
+func WithAllowedUpdateNamespaces(value []string) controllerOption {
+	return func(c *controllerConfig) {
+		c.allowedUpdateNamespaces = value
```
rbren, line 74: do we use `allowedNamespaces` as the default if `allowedUpdateNamespaces` is unspecified? I'm not seeing that logic.

I'd imagine something like
```go
func WithAllowedNamespaces(value []string) controllerOption {
  return func(c *controllerConfig) {
    c.allowedNamespaces = value
    if c.allowedUpdateNamespaces == nil {
      c.allowdedUpdateNamespaces = value
    }
  }
}
```

plugins/right-sizer/src/controller/controller.go
ivanfetch-fw: > I'm not sure I agree with this check - why not just do everything in `allowedUpdateNamespaces`?
Do you not agree with the option in its entirety (a separate namespace filter only applying to updating limits), or where the check is in the code?

> Also - don't we have to worry about the case where `allowedNamespaces` is empty (meaning all namespaces)?
Yes, I also realized this earlier today - fixed. Thanks for catching this!

plugins/right-sizer/src/controller/controller.go
ivanfetch-fw: The global namespaces aren't used as the default, only because the check of global namespaces happens earlier in logic and short-circuits the check of update-namespaces: IF the global namespaces is set, and a given namespace isn't in that list, it skips processing that pod, including caring about whether limits should be updated, Etc.

plugins/right-sizer/src/controller/controller.go
```diff
@@ -122,6 +132,12 @@ func NewController(stop chan struct{}, kubeClientResources util.KubeClientResour
 		o(cfg)
 	}
 
+	missingAllowedNamespaces := funk.RightJoinString(cfg.allowedNamespaces, cfg.allowedUpdateNamespaces)
+	if len(cfg.allowedNamespaces) > 0 && len(missingAllowedNamespaces) > 0 {
```
rbren, line 136: Should we be checking if len > 0 here? I think we'd always want to do this...

ivanfetch-fw: @rbren If no global namespaces were specified, and we always added any update namespaces to the global list, we would unintentionally limit namespaces for which action-items happen. By default an empty global namespaces list means action-items happen everywhere. I want to make sure this logic makes sense and I'm not missing your point, before I merge.

ivanfetch-fw: Robert approved this point in [this Slack thread](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1642113005253000?thread_ts=1642112819.251000&cid=CNQBGRGFN)

PR #582, opened 2022-01-14
Add admission version as flag

makoscafee:
**Why This PR?**
_a short description of why this PR is needed_

Fixes #

**Changes**
Changes proposed in this pull request:
*

**Checklist:**

* [ ] I have signed the CLA
* [ ] I have updated/added any relevant documentation related to this PR


plugins/admission/pkg/admission/admission.go
```diff
@@ -19,11 +19,14 @@ import (
 var organization string
 var hostname string
 var cluster string
+var version string
 
 func init() {
 	organization = os.Getenv("FAIRWINDS_ORGANIZATION")
 	hostname = os.Getenv("FAIRWINDS_HOSTNAME")
 	cluster = os.Getenv("FAIRWINDS_CLUSTER")
+	version = "1.0.0"
```
rbren, line 28: can we just get this from `version.txt` using https://pkg.go.dev/embed? You might need to add it to the dockerfile.

.circleci/config.yml
```diff
@@ -137,6 +137,7 @@ jobs:
             cd ../..
           done
       - run: cd plugins/opa ; go test ./... ; cd ../..
+      - run: cp plugins/admission/version.txt plugins/admission/pkg/admission/version.txt
```
rbren, line 140: are we able to keep it where it is, and make the embed relative? Or maybe we could put the embed in main.go?

rbren (commented):
Nice work!

I think we need to do the same for the OPA and Polaris admission reports (e.g. [here](https://github.com/FairwindsOps/insights-plugins/blob/main/plugins/admission/pkg/opa/opa.go#L16))

.circleci/config.yml
ivanfetch-fw: Go Embed doesn't allow periods in the (relative) file path. The only way  I've been able to embed version.txt from the top of the plugin's directory is to use a hard-link at the same directory as the .go file doing the embedding.
 However a hard-link may feel like a separate file until someone troubleshoots and realizes it is a hard-link (filesystem inode is the same).

Although the Dockerfile `COPY` directive works, sticking with that option will cause local admission builds to fail ... even though a local `cd admission && go build cmd/admission/main.go` is not a complete build because it lacks things like Trivy and Polaris.

How do you both feel about using a hard-link for version.txt?

rbren: Huh. I've never used hard-links before. Seems like a reasonable solution to me!

ivanfetch-fw: Unfortunately git sees a hard link as a separate file because it does not care about inode numbers.

ivanfetch-fw: After [additional conversation in Slack](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1646860188680969) I came up with another potential option to read version.txt without the need to copy that file in CI or docker builds - see [PR #625](https://github.com/FairwindsOps/insights-plugins/pull/625). We can talk about this more next week when Barnie’s back.

ivanfetch-fw: As [this alternative PR](https://github.com/FairwindsOps/insights-plugins/pull/625) has ben merged which accomplishes the same goals of this PR, I'll go ahead and close this PR.

PR #588, opened 2022-01-20
FWI-1261 - Remove OPA CRD

makoscafee:
This PR fixes issue [FWI-1261](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1261)

## Checklist
* [x] I have signed the CLA
* [x] I have updated/added any relevant documentation

### What's the goal of this PR?
Remove CRD dependency for `OPA` and instead use API to get all the checks
### What alternative solution should we consider, if any?

[Internal Ticket FWI-1261](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1261)

PR #596, opened 2022-01-25
Add `insightsinfo` function available in rego policies

ivanfetch-fw:
NOTE: Do not merge this until [this related API PR](https://github.com/FairwindsOps/Insights/pull/2725) has been included in a release.

## Checklist
* [x] I have signed the CLA

* Make a rego function `insights info` available to policies which allows policy writers to access information that has previously been specified in our instance yaml. The OPA, CI, and Admission plugins now pass this context to the rego engine.
  * The Insights context (from where rego has been run): CI, admission-controller, or the OPA plugin..
  * The cluster where a policy is running.
* End up resolving a bug that allowed failures in rego functions (such as our `Kubernetes` one) to silently pass that policy. For example, looking up an invalid APIGroup or Kind in a policy.
* Remove duplicate logging from functions that pass errors up the stack, where they are also logged. Plugin output now reads:
* ```
* time="2022-02-07T16:52:59Z" level=info msg="Starting OPA reporter"
time="2022-02-07T16:53:00Z" level=info msg="Found 2 checks"
time="2022-02-07T16:53:01Z" level=error msg="There were errors while processing OPA checks."
1 error occurred:
	* error while processing check instance /deployments: Error while processing check test: fairwinds:5: eval_builtin_error: insightsinfo: cannot return unknown Insights Info "context2"
```

### Output / Behavior

* As part of testing that the new `insightsinfo` function works in rego, I created a rego policy that sets the action-item description to `"description": sprintf("insights context %v and cluster %v", [insightsinfo("context"), insightsinfo("cluster")]),`, which creates an action-item with its description set to `insights context Agent and cluster us-east-1`

* If an unknown parameter is passed to the `insightsinfo` function via a rego policy, no action-item is created, but the OPA plugin Pod logs a warning in its final output:
```
time="2022-02-07T16:53:01Z" level=error msg="There were errors while processing OPA checks."
1 error occurred:
	* error while processing check instance /deployments: Error while processing check test: fairwinds:5: eval_builtin_error: insightsinfo: cannot return unknown Insights Info "context2"
```

* A failure when calling the `kubernetes` function **no longer silently fails and allows the policy** - the policy now fails and logs. This also keeps the admission controller from allowing resources due to these errors.
  * Example OPA plugin log for invalid use of `kubernetes()` -> `* error while processing check instance /deployments: Error while processing check test: fairwinds:5: eval_builtin_error: kubernetes: Error while getting data for apps/deployment2: no matches for kind "deployment2" in group "apps"`
  * Admission controller response when using an invalid parameter to the insightsinfo function
    * In rego, calling `insightsinfo("context2")`
    * The request is denied with: `Error from server: error when creating "manifest.yaml": admission webhook "insights.fairwinds.com" denied the request: fairwinds:5: eval_builtin_error: insightsinfo: cannot return unknown Insights Info "context2"`
  * Admission controller response when using an invalid Kind with the kubernetes function
    * In rego, calling `kubernetes("apps", "IncorrectKindGoesHere")`
    * The request is denied with: `Error from server: error when creating "manifest.yaml": admission webhook "insights.fairwinds.com" denied the request: fairwinds:5: eval_builtin_error: kubernetes: no matches for kind "IncorrectKindGoesHere" in group "apps"`

I added a single `insights info` function available to the rego engine, which accepts a single parameter representing which information to return from a new `InsightsInfo` struct.

I used a `multi error` package to return multiple errors while processing OPA checks, instead of returning the last error.

I thought about creating separate functions to return Insights Context, cluster, Etc. I think a single function is a strait-forward window to Insights via rego, similar to the existing `kubernetes` function. A single `insights info` function also eases exposing new information to rego in our future.

### CODE REVIEW

File: plugins/opa/pkg/opa/opa.go [hunk 1]
```diff
@@ -135,8 +135,8 @@ func processCheckTarget(ctx context.Context, check CustomCheck, checkInstance Cu
 	return actionItems, nil
 }
 
-func ProcessCheckForItem(ctx context.Context, check CustomCheck, instance CustomCheckInstance, obj map[string]interface{}, resourceName, resourceKind, resourceNamespace string) ([]ActionItem, error) {
-	results, err := runRegoForItem(ctx, check.Spec.Rego, instance.Spec.Parameters, obj)
+func ProcessCheckForItem(ctx context.Context, check CustomCheck, instance CustomCheckInstance, obj map[string]interface{}, resourceName, resourceKind, resourceNamespace string, insightsInfo *rego.InsightsInfo) ([]ActionItem, error) {
+	results, err := runRegoForItem(ctx, check.Spec.Rego, instance.Spec.Parameters, obj, insightsInfo)
```
rbren, line 139: great idea to pass this in as a struct!

rbren (approved):

ivanfetch-fw:
I've merged main back into this PR, and resolved conflicts from updates to PR #588  which changed some of the same code. I'm re-requesting review CC @rbren

plugins/opa/pkg/opa/opa.go
```diff
@@ -67,7 +67,7 @@ func processAllChecks(ctx context.Context, checkInstances []CheckSetting, checks
 			}
 		}
 	}
-	return actionItems, lastError
+	return actionItems, nil
```
ivanfetch-fw, line 70: @rbren I wanted to call out the reason I no longer return `lastErr` is to avoid the calling function and main.go propagating these errors as a panic.

rbren: 🤔 I think we should still pass the error back here, but not panic when it does return

rbren (approved):
Thanks Ivan!

rbren (approved):
Looks good Ivan!

I think maybe we should bump the minor versions on CI and Admission

PR #605, opened 2022-02-02
FWI-1338 - Update OPA plugin to process v 2.0 checks

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

* OPA, Admission, and CI Plugins:
  * Support V2 of CustomChecks (without an accompanying YAML instance or parameters).
* OPA Plugin:
  * Change the `EventType` to be the CustomCheck name for V1 checks, to match the same behavior for V2 checks, provide more troubleshooting context, and and resolve [bug FWI-1362](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1362). See also [this Slack comment from Robert](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1644271153009399?thread_ts=1644268298.475679&cid=CNQBGRGFN).
  * Processing of checks is no longer interrupted by a failure to list objects for one of the checks Kube targets. The remaining targets will be checked, and errors reflected in the plugin log.
    * Errors are limited to 5 per check, unless debug-logging is enabled. THIs avoids noisy logs for large clusters where I.E. rego might fail for all deployments.
  * Add command-line options to specify Kubernetes resource targets for V2 custom checks, and to enable debug logging.
* Admission Plugin:
  * The cluster name is now correctly available via the `insightsinfo("cluster")` rego function.
  * Processing of checks will now continue when there has been a failure, to collect and output all failure conditions. Multiple errors may be reflected in both admission webhook output and in plugin log output.
* CI Plugin:
  * Debug output can be enabled by setting the `LOGRUS_LEVEL` environment variable to `debug`.
  * Processing of checks will now continue when there has been a failure, to collect and output all failure conditions. Multiple errors may be reflected in plugin output.

[Internal Ticket FWI-1338](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1338)

[Internal Ticket FWI-1362](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1362)


plugins/opa/pkg/opa/opa.go
```diff
@@ -126,7 +141,29 @@ func processCheckTarget(ctx context.Context, check CustomCheck, checkInstance Cu
...
+	if err != nil {
+		return actionItems, err
+	}
+	gvr := mapping.Resource
+	list, err := client.DynamicInterface.Resource(gvr).Namespace("").List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, obj := range list.Items {
+		newItems, err := ProcessCheckForItemV2(ctx, check, obj.Object, obj.GetName(), obj.GetKind(), obj.GetNamespace(), &rego.InsightsInfo{InsightsContext: "Agent", Cluster: os.Getenv("FAIRWINDS_CLUSTER")})
```
rbren, line 166: is `FAIRWINDS_CLUSTER` set in both admission and agent?

rbren (commented):
This is looking great!

Will take another look once this gets rebase with Barnie's PR in

rbren:
Realizing this now - currently we use the instanceName as the action item `EventType`. Are we using checkName now?

plugins/ci/version.txt
```diff
@@ -1 +1 @@
-1.2.2
+1.2.3
```
rbren, line 1: We should call this 2.0 for consistency

plugins/opa/pkg/opa/opa.go
ivanfetch-fw: Yes, and I"ve (this week) fixed cluster being available in the `insights info` function via the admission plugin.

plugins/admission/pkg/opa/opa.go
```diff
@@ -22,46 +26,58 @@ func ProcessOPA(ctx context.Context, obj map[string]interface{}, resourceName, a
...
-					return opa.KubeTarget{
-						APIGroups: []string{splitValues[0]},
-						Kinds:     []string{splitValues[1]},
+	var allErrs error = nil
+	cluster := os.Getenv("FAIRWINDS_CLUSTER")
+	for _, check := range configuration.OPA.CustomChecks {
+		logrus.Debugf("Check %s is version %.1f\n", check.Name, check.Version)
+		switch check.Version {
+		case 1.0:
+			for _, instanceObject := range configuration.OPA.CustomCheckInstances {
```
rbren, line 35: there's a lot of nesting inside this `switch` - I would suggest breaking each case out into its own function.

plugins/opa/cmd/root.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	opa "github.com/fairwindsops/insights-plugins/opa/pkg/opa"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
```
rbren, line 19: We usually group all the third-party stuff in the middle of imports (i.e. above with logrus on line 14)

plugins/opa/cmd/root.go
```diff
@@ -0,0 +1,115 @@
...
+		viper.SetConfigFile(cfgFile)
+	} else {
+		// Find home directory.
+		home, err := os.UserHomeDir()
+		cobra.CheckErr(err)
+
+		// Search config in home directory with name ".opa" (without extension).
+		viper.AddConfigPath(home)
+		viper.SetConfigType("yaml")
+		viper.SetConfigName(".opa")
```
rbren, line 71: what does this do?

rbren (approved):
This looks great!

Approving as I think this is working well as-is, but left a couple suggestions for clean up. Mostly the big `switch` statements for v1 vs v2 - I think those would be easier to read and more reusable if we broke out each of the cases into its own function

plugins/admission/pkg/opa/opa.go
ivanfetch-fw: @rbren I've broken out those `cases into `ProcessOPAV1` and `ProcessOPAV2` functions, would you mind taking a look?

plugins/opa/cmd/root.go
ivanfetch-fw: It's Viper boilerplate to also use a config-file as optional input to specify command-line options. While I"m not processing it currently (have removed the option from showing up in `--help`), I left the code in here as an option, and because it was generated by Cobra.

ivanfetch-fw:
@rbren Would you mind confirming my version bump for OPA, Admission, and CI are appropriate? OPA and CI bumped from V1 to V2, but Admission is still at 1.1.1.

ivanfetch-fw:
I've pushed an update for the OPA Plugin that limits the number of logged errors to 5 per check, to avoid overwhelming logs in cases where rego fails for many Kube targets (like all deployments). IF plugin debug-logging is enabled, all errors will be logged.

 Here is an example of the error output - see the bottom, where the second primary error for the `test2` check has its sub-errors limited to 2 out of 5:

```
time="2022-03-01T19:28:19Z" level=error msg="There were errors while processing OPA checks."
2 errors occurred:
	* error while processing check test / instance deployments: 2 errors occurred:
	*   error while running rego for check test on item Deployment/cert-manager/cert-manager: fairwinds:5: eval_builtin_error: insightsinfo: cannot return unknown Insights Info "clusterX"
	*   no matches for kind "brokenCustomThing" in group "apps"


	* error while processing check test2: 5 errors occurred (only 2 are shown for brevity):
	*   error while running rego for check test2 on item DaemonSet/kube-system/kindnet: Error while evaluating query: fairwinds:5: eval_builtin_error: insightsinfo: cannot return unknown Insights Info "clusterX"
	*   error while running rego for check test2 on item Deployment/cert-manager/cert-manager: Error while evaluating query: fairwinds:5: eval_builtin_error: insightsinfo: cannot return unknown Insights Info "clusterX"

```

ivanfetch-fw:
See [this Slack thread](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1646246209549729?thread_ts=1646233454.703409&cid=CNQBGRGFN) for approval of changes to this PR after Robert initially approved it, and some discussion about what the CI and Admission plugin versions should be.

PR #608, opened 2022-02-04
add scan-all script

rbren:
This PR fixes issue #

## Checklist
* [x] I have signed the CLA

Have a way to scan all the current published versions of plugins in this repo

Added a shell script

Would be nice to e.g. run this on a cronjob


makoscafee (approved):
so this will be manually run locally?

PR #615, opened 2022-02-22
FWI-1442 - Plugin for gathering newest image versions

jdesouza:
This PR fixes #

## Checklist
* [ ] I have signed the CLA

[Internal Ticket FWI-1442](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1442)


plugins/trivy/cmd/trivy/main.go
```diff
@@ -100,4 +118,43 @@ func main() {
...
+				})
+			}
+		}
+	}
+	return newImagesToScan
+}
+
+func getNewestVersions(versionsChan chan newestVersions, ctx context.Context, img models.ImageReport) {
+	repo := strings.Split(img.Name, ":")[0]
+	tag := strings.Split(img.Name, ":")[1]
```
rbren, line 148: Would be good to check that `img.Name` is the right format so this doesn't crash. E.g.

```suggestion
  parts := strings.Split(img.Name, ":")
  if len(parts) != 2 {
    return
  }
  repo := parts[0]
  tag := parts[1]
```

plugins/trivy/pkg/image/minimize.go
```diff
@@ -24,14 +24,15 @@ func Minimize(images []models.ImageReport, lastReport models.MinimizedReport) mo
...
-			LastScan:       &timestamp,
+			ID:                 imageDetails.ID,
+			Name:               imageDetails.Name,
+			OwnerName:          imageDetails.OwnerName,
+			OwnerKind:          imageDetails.OwnerKind,
+			OwnerContainer:     imageDetails.OwnerContainer,
+			Namespace:          imageDetails.Namespace,
+			Report:             []models.VulnerabilityRefList{},
+			LastScan:           &timestamp,
+			RecommendationOnly: imageDetails.RecommendationOnly,
```
rbren, line 35: Before we merge this, we need to make sure the server is checking for this.

rbren (commented):
This looks great!

plugins/trivy/cmd/trivy/main.go
jdesouza: Good point

plugins/trivy/cmd/trivy/main.go
```diff
@@ -100,4 +118,47 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
+	logrus.Info("Finished writing file ", outputFile)
+}
+
+func getNewestVersionsToScan(ctx context.Context, imageWithVulns []models.ImageReport) []models.Image {
+	versionsChan := make(chan newestVersions, len(imageWithVulns))
+	for _, i := range imageWithVulns {
```
rbren, line 126: is it possible that there are duplicate images/repositories in here? we might be double-scanning if so

rbren (approved):
Thanks James!

LMK if it's possible to make this more efficient by only calling `getNewestVersions` once per repo, instead of once per image

PR #616, opened 2022-03-01
FWI-1499 - Update insights-plugins go.mod and import statements to include plugins sub-directory

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

Update go.mod and imports for our Insights Plugins to use the current path (which includes the `plugins` sub-directory). This will match [a previous repo update that changed the location of all plugins](https://github.com/FairwindsOps/insights-plugins/tree/f1a39d0), and allow external programs to import updated versions of plugin packages, such as [the Insights CLI](https://github.com/FairWindsOps/insights-cli).

* Update the `module` defined in each plugin's go.mod
* Update the existing `replace` in go.mod to reflect the new import path. This allows plugins to import one-another locally, facilitating development of plugins which depend on one-another.
* Use `go get ...` to re-download plugins for some plugins via their new path, and manually remove old plugin paths from go.mod and go.sum.
* Verify plugins compile locally and via Docker.

We could use go module `replace` to replace the old path with the new one, but I don't believe this is an appropriate or necessary use of replace.

[Internal Ticket FWI-1499](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1499)


rbren (approved):
I don't think this is controversial. Let's get it in before you have to rebase 😄

PR #623, opened 2022-03-09
FWI-1547 - Update example OPA policies to V2

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

Update our example OPA policies to the V2 format (stored in a separate directory for now).

* Replace Instance YAML with in-policy variables and matching the
  resource Kind in-policy.
* Update the ReadMe to reflect multiple options  for adding policies to Insights, and that policies contain configurable variables.
* Add a note to policies that match Kubernetes Kinds which are not processed by the OPA plugin by default, such as Secrets. This helps prompts users that the OPA plugin Kubernetes Targets need to be updated.
* Move some rules to the top of policies when they contain configurable variables, for better user visibility.
* Add case-insensitive logic when matching the "Kind" of a Kube object against multiple Kinds to which this policy should apply. This is meant to help avoid errors with capitalilzation in a longer list of Kinds I.E. `deployment` vs. `Deployment`.
* Add Kube manifests to each policy to help test it (see the `test-manifest` sub-directories).
* Further differentiate title/remediation of the two action-items returned by the `lb-vuln-cve-2020-8554` policy (loadbalancer IPs granted by the Kube provider vs. externalIPs defined in a Service).
* Other minor corrections to action items generated by policies.

[Internal Ticket FWI-1547](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1547)


plugins/opa/examples-v2/README.md
```diff
@@ -0,0 +1,18 @@
...
+* From the Insights user interface, see the `Create From Template` button under the Policy section.
+* Use the [Insights command-line](https://github.com/FairWindsOps/insights-cli) to upload policies using the `insights-cli policy sync` command.
+* Add a policy directly using the Insights API - for example: `curl --data-binary @filename.rego -X PUT -H "Content-type: text/plain" -H "Authorization: Bearer $FAIRWINDS_TOKEN" "https://insights.fairwinds.com/v0/organizations/$organization/opa/customChecks/$checkName?version=2.0"`
+  * Replace `$FAIRWINDS_TOKEN` with an organization API token.
+  * Replace `$organization` with an existing organization name in Insights.
+  * Replace `$checkName` with what you would like to name the new custom check in Insights. This name will show up in the action items created by the check.
+
+## Testing Policies
+
+Each policy contains a `test-manifest` sub-directory with Kubernetes manifests to help test and demonstrate what the policy should match.
```
rbren, line 18: is there a script or anything to run the tests?

I wonder if we should call them something like `deployment.fail.yaml` (so we can add a `deployment.pass.yaml` in the future to make sure the policy isn't catching the wrong things)

plugins/opa/examples-v2/annotation-required/test-manifest/deployment.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: policy-test
+  namespace: default
+  annotations:
+    # THis policy requires the below annotation.
```
rbren, line 7:
```suggestion
    # This policy requires the below annotation.
```

rbren (approved):

plugins/opa/examples-v2/README.md
ivanfetch-fw: For now these files all represent a way to speed coming up with something that will match the policy. I hesitate to name the files in a pass/fail way yet, because some of them require uncommenting a stanza (depending on whether the policy is blocking or requiring a thing).

PR #625, opened 2022-03-10
FWI-1484 - Fix plugin versions using embed package

ivanfetch-fw:
This PR is an alternate possibility for [PR 582](https://github.com/FairwindsOps/insights-plugins/pull/582) that avoids copying plugin version.txt files within the pkg sub-directory, and breaking local plugin builds.

## Checklist
* [x] I have signed the CLA

Submit the plugin version, taken from the version.txt file, in submissions to the Insights API, including plugin packages used by other plugins, such as Admission using OPA / Polaris.

* Added an `embed.go` to the root of each plugin, which must be located at the same location as the version.txt file we use to track the plugin version.
* Import the new package, using an alias: `admission version "GitHub.com/fairwindsops/insights-plugins/plugins/admission`
* Use these packages to obtain current plugin versions, such as `admission version.String()` or `opaversion.String()`.

[This Slack thread](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1646860188680969) contains discussion about other options.
* Using a hard link to mirror the version.txt file within pkg sub-directories doesn't work because git doesn't preserve hard links.
* Copying version.txt within pkg sub-directories (via the Dockerfile) does work, but it complicates / breaks a local `go build`, and requires the Admission Dockerfile to copy version.txt in place for other plugins (OPA / Polaris).

[Internal Ticket FWI-1484](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1484)


plugins/admission/pkg/admission/admission.go
```diff
@@ -57,7 +58,7 @@ func SendResults(reports []models.ReportInfo, token string) (passed bool, warnin
 
 	req.Header.Set("Content-Type", w.FormDataContentType())
 	req.Header.Set("Authorization", "Bearer "+token)
-
+	req.Header.Set("X-Fairwinds-Admission-Version", admissionversion.String())
 	for _, report := range reports {
 		req.Header.Set("X-Fairwinds-Report-Version-"+report.Report, report.Version)
```
rbren, line 63: I think this `report.Version` gets set somewhere in the admission package - are we able to update OPA here?

ivanfetch-fw: Yes - I've just pushed an update that adds `embed.go` to the OPA plugin and uses it within the Admission plugin / opa package for the OPA report.

Once we verify we like how this looks I think we can do the same for other reports that Admission uses.

plugins/admission/Dockerfile
```diff
@@ -13,12 +13,14 @@ COPY opa opa
 WORKDIR /go/src/github.com/fairwindsops/insights-plugins/plugins/admission
 
 COPY ./admission/go.* ./
+COPY ./admission/version.txt .
+COPY ./admission/embed.go .
+COPY ./admission/pkg pkg
+COPY ./admission/cmd cmd
```
rbren, line 19: I would keep the `COPY` statements below the `go mod download` for better caching.

rbren (commented):
This strategy looks great to me. Thanks for working through it Ivan and Barnie!

plugins/admission/Dockerfile
ivanfetch-fw: I've moved these back where they were. I had previously needed to copy some files before `go mod download` to avoid failure to find our new version pkg, but that is now solved by a `replace ...admission => ./` in the Admission go.mod so it doesn't look externally for the version pkg.

ivanfetch-fw:
> This strategy looks great to me. Thanks for working through it Ivan and Barnie!

@rbren / @makoscafee I'd like to call out that importing the new `version` pkg from `embed.go` is not able to include the word `version`; it uses the top-level plugin path in the import. That looks like:
```
import (
  admission version "GitHub.com/fairwindsops/...../admission"
)
```

I have tried to improve on this, but naming the package something more descriptive in the import would require the package be located in a different sub-directory ... which defeats the purpose of having `embed.go` located with version.txt.

I want to make sure you're aware of / ok with this?

rbren (approved):

ivanfetch-fw:
> > This strategy looks great to me. Thanks for working through it Ivan and Barnie!
> 
> @rbren / @makoscafee I'd like to call out that importing the new `version` pkg from `embed.go` is not able to include the word `version`; it uses the top-level plugin path in the import. That looks like:
> 
> ```
> import (
>   admissionversion "GitHub.com/fairwindsops/...../admission"
> )
> ```
> 
> I have tried to improve on this, but naming the package something more descriptive in the import would require the package be located in a different sub-directory ... which defeats the purpose of having `embed.go` located with version.txt.
> 
> I want to make sure you're aware of / ok with this?

We [covered this in Slack](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1647372230750599?thread_ts=1647369116.134959&cid=CNQBGRGFN):
RobertB: `:nod: yeah I think that's fine` At 3:28 PM yesterday

PR #627, opened 2022-03-11
set up scan schedule

rbren:
This PR fixes #

## Checklist
* [x] I have signed the CLA

Get notified when new vulnerabilities appear in our latest images

Added a script to run every day and scan for vulns


.circleci/scripts/install-trivy.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+set -eo pipefail
+
+curl -L https://github.com/aquasecurity/trivy/releases/download/v0.23.0/trivy_0.23.0_Linux-64bit.tar.gz > trivy.tar.gz
```
makoscafee, line 4: we are in version `v0.24.` now

makoscafee (approved):
LGTM

.circleci/scripts/install-trivy.sh
rbren: Good catch!

PR #637, opened 2022-03-25
FWI-1663 - Trivy report: delete old image recommendations

jdesouza:
This PR fixes #

## Checklist
* [ ] I have signed the CLA

[Internal Ticket FWI-1663](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-1663)


rbren (commented):
What happens if:
* we scan image A
* we get a recommendation for image A
* image A gets deleted from the cluster

I think currently we'd keep the recommendation?

We need to make sure we delete _both_ the scan data and the recommendation data if image A isn't returned from `getAllImages)

plugins/trivy/cmd/trivy/main.go
```diff
@@ -141,15 +146,30 @@ func getImagesToScan(images []models.Image, lastReportImages []models.ImageDetai
...
+		if report.RecommendationOnly {
+			// Add old recommendations only if we have not scanned for new recommendations
+			if _, found := newRecommendations[reportSha]; !found {
+				imagesToKeep = append(imagesToKeep, report)
+			}
+			// Add old recommendations only if we have the images they are for still running in the cluster
+			parts := strings.Split(report.Name, ":")
+			if _, found := clusterImagesMap[parts[0]]; found {
+				imagesToKeep = append(imagesToKeep, report)
+			}
```
rbren, line 168: it seems like we could possibly add it twice here...

rbren (approved):

PR #655, opened 2022-05-17
Admission Webhook mutation

makoscafee:
This PR fixes #

## Checklist
* [x] I have signed the CLA

The goal is to use Polaris mutation option to mutate failed resources.


plugins/admission/pkg/polaris/polaris.go
```diff
@@ -38,3 +42,31 @@ func GetPolarisReport(ctx context.Context, config polarisconfiguration.Configura
...
+	var controller kube.GenericResource
+	var err error
+	if kind == "Pod" {
+		pod := corev1.Pod{}
+		err := decoder.Decode(req, &pod)
+		if err != nil {
+			return nil, err
+		}
+		if len(pod.ObjectMeta.OwnerReferences) > 0 {
+			logrus.Infof("Allowing owned pod %s/%s to pass through webhook", pod.ObjectMeta.Namespace, pod.ObjectMeta.Name)
```
ivanfetch-fw, line 56: Does this mean that stand-alone pods would not be mutated, or is there additional significance (like admission-control)?

plugins/admission/pkg/polaris/polaris.go
```diff
@@ -38,3 +42,31 @@ func GetPolarisReport(ctx context.Context, config polarisconfiguration.Configura
...
+			return nil, nil
+		}
+		controller, err = kube.NewGenericResourceFromPod(pod, pod)
+	} else {
+		controller, err = kube.NewGenericResourceFromBytes(req.Object.Raw)
+	}
+	if err != nil {
+		return nil, err
+	}
+	// TODO: consider enabling multi-resource checks
```
ivanfetch-fw, line 66: I'm curious to know more about what a multi-resource check would be - perhaps a deployment and accompanying HPA?

ivanfetch-fw (commented):
I've added a few curiosity questions. As I continue to read the Polaris code that this PR leverages, I'm a bit stuck on how `ResourceProvider` relates to `Resource`, for example the Polaris `func ApplyAllSchemaChecks`. Perhaps we can talk through this a bit next week.

plugins/admission/pkg/polaris/polaris.go
makoscafee: Actually that `TODO` should not be there... i do not think there is a point we would have multiple resources validated at once.

plugins/admission/pkg/polaris/polaris.go
makoscafee: I think that is the opposite... we do not process/mutate pods that have OwnerReference, meaning they are not standalone because they get covered in the else statement.

ivanfetch-fw (approved):
While I don't yet track the upstream Polaris functions this code uses, this changes to fulfill a mutating admission request look good to me.

PR #677, opened 2022-07-21
FWI-940 - OPA should not panic on missing kind manifest files

makoscafee:
This PR fixes #

## Checklist
* [ ] I have signed the CLA

[Internal Ticket FWI-940](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-940)


ivanfetch-fw (approved):
@makoscafee How does CI behave when continuing to process a manifest which is missing a Kind?

plugins/ci/pkg/ci/opa.go
```diff
@@ -56,14 +56,20 @@ func (ci CIScan) ProcessOPA(ctx context.Context) (models.ReportInfo, error) {
...
+					for _, node := range nodes {
+						nodeMap := node.(map[string]interface{})
+						files = append(files, nodeMap)
+					}
+				} else {
+					files = append(files, yamlNode)
 				}
 			} else {
+				logrus.Warn("Manifest is missing field Kind")
 				files = append(files, yamlNode)
```
rbren, line 73: If Kind is missing, we should skip this file.

makoscafee: aah gotcha!

makoscafee:
> @makoscafee How does CI behave when continuing to process a manifest which is missing a Kind?

I think the idea is to ignore them, and not process them at all.

PR #682, opened 2022-08-01
improve-image-tag-recommendation

vitorvezani:
This PR fixes #

## Checklist
* [X] I have signed the CLA

Improve image tags recommendations


plugins/trivy/pkg/image/newest_images.go
```diff
@@ -54,22 +60,47 @@ func fetchTags(ctx context.Context, imageRepoName string) ([]string, error) {
...
+	for _, tag := range suggestedTags {
+		v, err := semver.NewVersion(tag)
+		if err != nil {
+			logrus.Infof("could not parse tag %s: %v", tag, err)
+			continue
+		}
+
+		cPre := curVersion.Prerelease()
+		vPre := v.Prerelease()
+		isKnownPreRelease := lo.Contains(knownPreReleases, cPre)
```
rbren, line 85: I don't think we need to check `knownPreReleases` - just that the two match

jdesouza (approved):

plugins/trivy/pkg/image/newest_images.go
vitorvezani: If we remove that, `v0.0.16-alpine` will not be recommended for `v0.0.14-beta1`

vitorvezani: fixed, I guess that was intended

plugins/trivy/pkg/image/version.go
```diff
@@ -0,0 +1,34 @@
...
+	tags := []string{}
+	for _, v := range s {
+		tags = append(tags, v.Original())
+	}
+	return tags
+}
+
+// Sort sorts the given slice of Version
+func Sort(versions []*semver.Version) {
+	sort.Sort(Versions(versions))
```
rbren, line 33: Why the `Versions()` part of this?

rbren (approved):

plugins/trivy/pkg/image/version.go
vitorvezani: `sort.Sort` requires a [sort.Interface](https://pkg.go.dev/sort#Interface) implementation, `Versions` is the struct that implements that

PR #696, opened 2022-08-30
FWI-2695 - Plugin - Add NamespaceLabels to admission controller payload

vitorvezani:
This PR fixes #
- Minor refactor to favor dependency injection over global package level state
- Adds `namespaceMetadata` field into `metadata` report

## Checklist
* [X] I have signed the CLA

[Internal Ticket FWI-2695](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-2695)


plugins/admission/pkg/admission/validate.go
```diff
@@ -135,20 +160,24 @@ func (v *Validator) Handle(ctx context.Context, req admission.Request) admission
...
-	report := models.ReportInfo{
-		Report:  "metadata",
-		Version: "0.1.0",
-	}
-	var err error
-	report.Contents, err = json.Marshal(&req.AdmissionRequest)
-	return report, err
+type MetadataReport struct {
+	admissionv1.AdmissionRequest
+	NamespaceObject map[string]any `json:"namespaceObject,omitempty"`
```
vitorvezani, line 165: Any suggestion on the name here? `namespace` is already taken by the namespace string value in `admissionv1.AdmissionRequest`

plugins/admission/pkg/admission/validate.go
```diff
@@ -135,20 +160,24 @@ func (v *Validator) Handle(ctx context.Context, req admission.Request) admission
...
-func processInputYAML(ctx context.Context, configurationObject models.Configuration, input []byte, decodedObject map[string]interface{}, token, name, namespace, kind, apiGroup string, metaReport models.ReportInfo) (bool, []string, []string, error) {
-	reports := []models.ReportInfo{
-		metaReport,
-	}
+func getRequestReport(req admission.Request, namespaceObject map[string]any) (models.ReportInfo, error) {
+	metadataReport := MetadataReport{AdmissionRequest: req.AdmissionRequest, NamespaceObject: namespaceObject}
+	contents, err := json.Marshal(&metadataReport)
+	return models.ReportInfo{
+		Report:   "metadata",
+		Version:  "0.2.0",
```
vitorvezani, line 173: I bump the version to `0.2.0` as we are adding a new field to it... or `0.1.1` would be enough?

plugins/admission/cmd/admission/main.go
```diff
@@ -33,21 +34,14 @@ func exitWithError(message string, err error) {
 	}
 }
 
-var handler fadmission.Validator
```
vitorvezani, line 36: I removed package vars usage and env vars usage scattered over the app, instead the functions now depend on the `InsightsConfig` struct, I think this will make test easier

plugins/admission/pkg/admission/validate.go
rbren: I don't think we really use this version, so this is fine

plugins/admission/pkg/admission/validate.go
ivanfetch-fw: A field name of `NamespaceMetadata` might be slightly more informative, but I don't feel strongly about the suggestion.

ivanfetch-fw (approved):
Awesome - besides the addition of namespace metadata which we can act on, I appreciate the new Insights configuration struct that sources environment variables in a single place.

plugins/admission/pkg/admission/validate.go
vitorvezani: Thanks, Ivan... I like `NamespaceMetadata` better, will change!

vitorvezani: Done

PR #697, opened 2022-09-02
FWI-2733 - Update Insights plugins to build for the arm64 platform

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

* Build plugins for the arm64 architecture, using goreleaser
* Fix the failing Kubernetes E2E tests
  * Get IPv4 IP for use by insights-uploader instead of potentially the IPv6 address
  * Temporarily disable right-sizer (addressed later / separately)
* Bump plugins to Go 1.19.1

* Add goreleaser configuration for each plugin, to build (if applicable) and create Docker images.
* Update plugin Dockerfiles to work with goreleaser, and download the correct OS and architecture of supporting binaries such as `Kubectl`, `Trivy`, Etc.
* Update the `template` public directory to include a starter Dockerfile and goreleaser configuration.
* Add a goreleaser wrapper script to create a temporary git tag from a plugin `version.txt`, and use `envsubst` to only push some Docker manifests for the git main branch.
  * Only versioned (x.y.z, x.y, x) and latest Docker manifests will be pushed for the main git branch.
  * Only a Docker manifest named after the git branch will be pushed for non-main; feature branches.

The built-in Docker `ARG`s for architecture and OS are passed in at built-time. These are now used in place of hard-coded operating system and architecture when downloading supporting binaries like Trivy or kubectl.

For Trivy, an if statement is used to map `64bit` to `amd64` to download the correct architecture.

The `envsubst` tool is used to substitute some environment variables in the goreleaser configuration, to vary when some Docker manifests are pushed.

[Internal Ticket FWI-2733](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-2733)


ivanfetch-fw:
I'm revisiting using GoReleaser for these builds, so have marked this PR as draft - stay tuned. :)

.circleci/config.yml
```diff
@@ -71,27 +75,20 @@ executors:
   ci-images:
     docker:
       - image: quay.io/reactiveops/ci-images:v11-stretch
-  ci-images-large:
+  goreleaser:
     resource_class: large
```
rbren, line 79:
```suggestion
```

rbren (approved):

.circleci/config.yml
ivanfetch-fw: @rbren It looks like you're suggesting to remove the large resource-class - we may need this to help goreleaser have CPU/memory to build all of the plugins.

rbren: I added the large resource class a while ago to help speed up the build. It's possible we still need it, but let's see how it runs without

rbren: (we can do that in a follow-up PR if you prefer)

ivanfetch-fw: @rbren I put this executor size back because the builds ran out of memory.

ivanfetch-fw:
Holding off on merging until we resolve the goreleaser docker image using Go 1.18.4, vs. latest 1.18.6. We may also attempt an update to latest goreleaser + Go 1.19.X.

PR #701, opened 2022-09-22
[prometheus] Better pod matching

rbren:
This PR improves the pod matching scheme for prometheus

## Checklist
* [x] I have signed the CLA


jdesouza (approved):

plugins/prometheus/pkg/data/data.go
```diff
@@ -171,19 +168,18 @@ func GetMetrics(ctx context.Context, dynamicClient dynamic.Interface, restMapper
 	if err != nil {
 		return nil, err
 	}
-	workloadMap := make(map[string]controller.Workload)
+	workloadMap := make(map[string]*controller.Workload)
 	for _, workload := range workloads {
 		for _, pod := range workload.Pods {
-			workloadMap[fmt.Sprintf("%s/%s", pod.GetNamespace(), pod.GetName())] = workload
+			workloadMap[fmt.Sprintf("%s/%s", pod.GetNamespace(), pod.GetName())] = &workload
```
jdesouza, line 174: @rbren I think this will return same pointer address all the time

jdesouza: Maybe we don't need this PR anymore 😆

rbren: 😬 I always mess this up

PR #704, opened 2022-10-05
Added ignore systems accounts options

jdesouza:
This PR fixes #

## Checklist
* [ ] I have signed the CLA

### DIFF
--- a/plugins/admission/pkg/admission/validate.go
+++ b/plugins/admission/pkg/admission/validate.go
@@ -3,10 +3,12 @@ package admission
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"os"
 	"strings"
 
+	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
 	admissionv1 "k8s.io/api/admission/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


plugins/admission/pkg/models/models.go
```diff
@@ -35,7 +35,8 @@ type Configuration struct {
 		CustomChecks         []opa.OPACustomCheck
 		CustomCheckInstances []opa.CheckSetting
 	}
-	Polaris *polarisconfiguration.Configuration
+	Polaris               *polarisconfiguration.Configuration
+	IgnoreServicesAccount []string
```
vitorvezani, line 39: Shouldn't it be called `IgnoreUsernames`?

jdesouza: Not sure which one is more clear

rbren: Yeah I think `Usernames` is better--service accounts are a type of user.

rbren: I'm not sure this should go in the config. That gets stored in the API IIRC. I was thinking it would be a flag in the helm chart (e.g. an environment variable or command-line flag).

Maybe this is better though? I'm not sure. Lemme think about it

ivanfetch-fw (approved):
While I did not test this by running admission-controller, the implementation looks good. Thanks James.

ivanfetch-fw (approved):
I know you're updating version.txt and changelog.md again, but I'm re-approving this per your request in Slack.

plugins/admission/pkg/admission/validate.go
```diff
@@ -95,12 +97,16 @@ func (v *Validator) InjectConfig(c models.Configuration) error {
 
 func (v *Validator) handleInternal(ctx context.Context, req admission.Request) (bool, []string, []string, error) {
 	var decoded map[string]interface{}
+	username := req.UserInfo.Username
+	if lo.Contains(v.iConfig.IgnoreUsernames, username) {
+		msg := fmt.Sprintf("Insights admission controller is ignoring service account %s.", username)
```
vitorvezani, line 102: `ignoring service account` → `ignoring username`?

PR #710, opened 2022-10-12
Add script to check for latest versions being used

rbren:
This PR create a script which will check if all the references in the insights-agent chart are up-to-date.

The script will run every night, and send a slack message if something is off.

## Checklist
* [x] I have signed the CLA


scripts/check-all-versions.sh
```diff
@@ -0,0 +1,92 @@
...
+for proj in ${plugin_projects[@]}; do
+  rewrites[$proj]=$proj
+done
+rewrites["aws-costs"]="awscosts"
+rewrites["falco-agent"]="falco"
+rewrites["kube-bench-aggregator"]="kube-bench.aggregator"
+rewrites["prometheus"]="prometheus-metrics"
+
+for proj in ${plugin_projects[@]}; do
+  latest_versions[$proj]=$(cat ../plugins/$proj/version.txt)
```
ivanfetch-fw, line 44: Does the insights-plugins repo need to be re-cloned? It looks like that copy is not used.

ivanfetch-fw (approved):
One non-blocking comment/question in the script, but this looks awesome.

scripts/check-all-versions.sh
```diff
@@ -0,0 +1,92 @@
+#! /bin/bash
+set -eo pipefail
+
+RED='\033[0;31m'
+NO_COLOR='\033[0m'
+
+if mkdir tmp_repos; then
+  cd tmp_repos
+  git clone https://github.com/FairwindsOps/charts
+  git clone https://github.com/FairwindsOps/insights-plugins
```
rbren, line 10:
```suggestion
```

PR #714, opened 2022-10-18
Use the `imageID` to get container images instead of `image`

ivanfetch-fw:
Sometimes `image` only contains a SHA. This change aims to provide an image name in more cases.
If `imageID` cannot be split by an at-sign (@), the original `image` is used as a fallback.
## Checklist
* [x] I have signed the CLA

Provide a name for Trivy image findings in cases where the `containerStatuses.*.image` field only contains a SHA.

Use the `containerStatuses.*.imageID` field instead, splitting it by an at-sign (@) to remove the SHA, and removing the optional `DockerPullable` prefix.

We could attempt to use the Pod `spec.Containers.*.image` field however that array can be in a different order than `containerStatuses`.

### DIFF
--- a/plugins/trivy/pkg/image/getimages.go
+++ b/plugins/trivy/pkg/image/getimages.go
@@ -96,8 +96,16 @@ func GetImages(ctx context.Context) ([]models.Image, error) {
 		}
 
 		for _, containerStatus := range pod.Status.ContainerStatuses {
+			var imageName string
+			imageIDFields := strings.SplitN(containerStatus.ImageID, "@", 2) // split image from SHA
+			if len(imageIDFields) < 2 || imageIDFields[0] == "" {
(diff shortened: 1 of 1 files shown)


plugins/trivy/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.23.3
```
vitorvezani, line 3: The new version should probably be `0.23.0` right?

plugins/trivy/version.txt
```diff
@@ -1 +1 @@
-0.22.3
+0.23.3
```
vitorvezani, line 1: Same as the first comment of new version being 0.23.0

plugins/trivy/CHANGELOG.md
ivanfetch-fw: Doe - thank you!

vitorvezani (approved):

PR #724, opened 2022-11-17
FWI-3154 - Update the CI plugin to succeed on content errors and report them in action items

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

* Continue processing reports for errors that are "soft," such as one failure out of many `helm template` commands, downloading images to be scanned, or processing OPA policies across multiple manifests.
* Report CI errors as action items, via a new `scan-errors` report type.
* Report-processing functions now return no report as `nil`, to differentiate whether  a report has data, since these functions can now return both a report and errors encountered during report processing.
* Provide a `CI_BASE_PATH` environment variable to override the default base path of `/app/repository` - useful when testing locally.

* Return a new `ScanErrorsReportResult` type for CI errors that we want to reflect as action items. This type satisfies the Go Error interface, and allows specifying additional context and metadata at the time the error happened.
* Update the `CIScan.ProcessRepository()` function to add the new `scan-errors` report type to the slice of reports, instead of short-circuiting and failing a CI scan when errors are encountered.
* Update error-processing throughout the CI plugin, to return the above `ScanErrorsReportResult` type instead of a traditional Error.

[Internal Ticket FWI-3154](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-3154)


plugins/ci/pkg/ci/helm.go
```diff
@@ -136,26 +161,50 @@ func handleLocalHelmChart(helm models.HelmConfig, baseRepoFolder, tempFolder str
...
+		}
 	}
 
 	var helmValuesFileArgs []string
 	for _, vf := range helmValuesFiles {
-		helmValuesFileArgs = append(helmValuesFileArgs, "-f", filepath.Join(repoPath, vf))
+		if filepath.IsAbs(vf) {
+			helmValuesFileArgs = append(helmValuesFileArgs, "-f", vf)
+		} else {
+			helmValuesFileArgs = append(helmValuesFileArgs, "-f", filepath.Join(repoPath, vf))
```
rbren, line 194: what's this?

plugins/ci/pkg/ci/polaris.go
```diff
@@ -9,22 +10,22 @@ import (
...
-		return report, err
-	}
 	polarisVersion, err := commands.Exec("polaris", "version")
 	if err != nil {
-		return report, err
+		return nil, fmt.Errorf("unable to get polaris version: %v: %v", err, polarisVersion)
 	}
 	report.Version = strings.Split(polarisVersion, ":")[1]
-	return report, nil
+	// Scan with Polaris
```
rbren, line 23: This comment is probably unnecessary. Great use of comments overall though!

rbren (approved):
Awesome work Ivan! This is super thorough. I only had a couple notes/questions.

plugins/ci/pkg/ci/helm.go
```diff
@@ -101,9 +103,16 @@ func handleRemoteHelmChart(helm models.HelmConfig, tempFolder string, configFold
...
-	_, err := commands.ExecWithMessage(exec.Command("helm", "repo", "add", repoName, helm.Repo), "Adding chart repository: "+repoName)
+	output, err := commands.ExecWithMessage(exec.Command("helm", "repo", "add", repoName, helm.Repo), "Adding chart repository: "+repoName)
 	if err != nil {
-		return err
+		return models.ScanErrorsReportResult{
+			ErrorMessage: fmt.Sprintf("%v: %s", err, output),
+			ErrorContext: fmt.Sprintf("adding helm repository %q", helm.Repo),
+			Kind:         "HelmChart",
+			ResourceName: helm.Name,
+			Filename:     helm.Name,
```
ivanfetch-fw, line 113: Is there something else we should use as the CI filename for scan-errors associated with a Helm chart? Here I use the name of the fairwinds-insights.yaml Helm definition.

plugins/ci/pkg/ci/trivy.go
```diff
@@ -131,12 +140,23 @@ func downloadMissingImages(folderPath string, dockerImages []trivymodels.DockerI
...
+		}
+	}
+	if len(allErrs.Errors) > 0 {
+		return ciutil.ReverseMap(refLookup), models.ScanErrorsReportResult{
+			ErrorMessage: allErrs.Error(), // keep multiple errors combined into a single error / action item
+			ErrorContext: "downloading missing images to be scanned by trivy",
+			Kind:         "InternalOperation",
+			Remediation:  "Please inspect the action item description for signs of why downloading images may have failed.",
+			ResourceName: "DownloadMissingImages",
+			Filename:     filepath.Clean(folderPath), // clean() removes potential trailing slash
```
ivanfetch-fw, line 158: Is it reasonable to use the general images directory as specified in fairwinds-insights.yaml, as the CI filename for errors encountered when downloading missing images?

plugins/ci/pkg/ci/ci.go
```diff
@@ -549,20 +554,33 @@ func readConfigurationFromReader(configHandler io.Reader) (*models.Configuration
...
-		return nil, fmt.Errorf("Error while processing helm templates: %v", err)
+		scanErrorsReportProperties.AddScanErrorsReportResultFromError(err)
 	}
 
 	err = ci.CopyYaml()
 	if err != nil {
-		return nil, fmt.Errorf("Error while copying YAML files: %v", err)
+		scanErrorsReportProperties.AddScanErrorsReportResultFromError(models.ScanErrorsReportResult{
+			ErrorMessage: err.Error(),
+			ErrorContext: "copying yaml files to configuration directory",
```
ivanfetch-fw, line 568: I've made a code readability choice to supply additional error context year, instead of each `if err != nil` inside the CopyYAML function.

plugins/ci/pkg/ci/ci.go
```diff
@@ -549,20 +554,33 @@ func readConfigurationFromReader(configHandler io.Reader) (*models.Configuration
...
+			Kind:         "InternalOperation",
```
ivanfetch-fw, line 569: I specify the required `Kind` field as `InternalOperation` often, because a lot of errors don't map back to a real Kube resource.

plugins/ci/pkg/ci/trivy.go
```diff
@@ -20,40 +20,47 @@ import (
...
 		return nil, err
 	}
 
 	filenameToImageName, err := downloadMissingImages(ci.config.Images.FolderName, dockerImages, manifestImages, ci.config.Options.RegistryCredentials)
 	if err != nil {
-		return nil, err
+		allErrs = multierror.Append(allErrs, err)
 	}
 
 	allImages, err := mergeImages(ci.config.Images.FolderName, dockerImages, manifestImages, filenameToImageName)
```
ivanfetch-fw, line 38: Should a failed call to `MergeImages` fail the trivy scan completely, or could the scan potentially continue?

plugins/ci/pkg/ci/ci.go
```diff
@@ -424,7 +424,12 @@ func getConfigurationForClonedRepo() (string, string, *models.Configuration, err
 		return "", "", nil, errors.New("IMAGE_VERSION environment variable not set")
 	}
 
-	basePath := filepath.Join("/app", "repository")
+	var basePath string = os.Getenv("CI_BASE_PATH")
```
ivanfetch-fw, line 427: This helps run the CI plugin outside of its typical docker container, for development / testing.

plugins/ci/pkg/ci/tfsec.go
```diff
@@ -10,92 +10,98 @@ import (
...
-			return models.ReportInfo{}, false, err
+			allErrs = multierror.Append(allErrs, fmt.Errorf("unable to process path %q: %v", terraformPath, err))
 		}
 		if len(results) > 0 {
 			reportProperties.Items = append(reportProperties.Items, results...)
 		}
 	}
+	var allErrsCombined error = nil
+	if len(allErrs.Errors) > 0 { // keep the multierror from becoming individual action items
+		allErrsCombined = fmt.Errorf("%v", allErrs.Error()) // return string representation from the multierror
```
vitorvezani, line 44: In some other places, you have used `allErrs.ErrorOrNil()`, wasn't possible to do it here?

vitorvezani: ohh, nvm I think I got it

plugins/ci/pkg/models/scan-errors.go
```diff
@@ -0,0 +1,143 @@
...
+// This satisfies the GO Error interface, and provides additional error context
+// to be included in scan-errors report action items.
+type ScanErrorsReportResult struct {
+	// IF adding a field to this struct, also update the FillUnsetFields
+	// receiver below!
+	Kind         string  `json:"kind"`
+	ResourceName string  `json:"resourceName"`
+	ErrorMessage string  `json:"errorMessage"` // error message returned during a scan
+	ErrorContext string  `json:"errorContext"` // where were we / what was happening when the error occurred
+	Filename     string  // filename being scanned that relates to this error
```
vitorvezani, line 25: Looks like it is missing the `json` tag for `Filename`

plugins/ci/pkg/ci/ci.go
vitorvezani: You could possibly move this to `(c *Configuration) SetDefaults()` that is called for both `auto-scan` and on `CI runners` - if that makes sense to do so...

vitorvezani (changes requested):
Great changes Ivan, I think the only required change is regarding the [Filename json tag missing](https://github.com/FairwindsOps/insights-plugins/pull/724/files#diff-67c79367322233097983f8c2eda409dc770a14b5bf013c5caccc8e9198892407R25)

plugins/ci/pkg/ci/ci.go
ivanfetch-fw: That relies on the Insights API, correct?

plugins/ci/pkg/models/scan-errors.go
ivanfetch-fw: Ah, it seemed to use the member name by default, but I've added the tag for clarity. Thanks.

ivanfetch-fw:
> Great changes Ivan, I think the only required change is regarding the [Filename json tag missing](https://github.com/FairwindsOps/insights-plugins/pull/724/files#diff-67c79367322233097983f8c2eda409dc770a14b5bf013c5caccc8e9198892407R25)

Thanks @vitorvezani 

Did you also see the comments I left in the PR, calling out a few things? One example is [this question about whether we should continue if merging images fails](https://github.com/FairwindsOps/insights-plugins/pull/724#discussion_r1061837431).

plugins/ci/pkg/models/scan-errors.go
vitorvezani: ohh, I thought that it would use pascal case by default if not specified... that is why I mentioned 😅

plugins/ci/pkg/ci/ci.go
vitorvezani: nvm, I've just checked and doesn't really make sense since the default value for `basepath` is different when using `auto-scan` and `CI`

- default `basePath` on auto-scan is `/app/repository`
- default `basePath` for CI is `.`

So we can't really unify the code that sets the `basepath`...

vitorvezani (approved):

plugins/ci/pkg/ci/trivy.go
vitorvezani: Seems like the only thing that may fail is walking through folders, I think it is doable to partially process using images that were processed before the error occurs...

plugins/ci/pkg/ci/helm.go
vitorvezani: Maybe having the `path` as filename would be useful?

plugins/ci/pkg/ci/trivy.go
vitorvezani: It could be either an image error or a writing error, ideally I would use `fairwinds-insights.yaml` for `dockerImages` downloading errors, however, as we are combining multiple errors into one AI there is not really a right answer to the filename.

I guess `folderPath` is reasonable (more so when it is a writing issue)

vitorvezani:
@ivanfetch-fw, just did... I think you have made good judgment calls, I left some thoughts, but none are really any definition or request for change

plugins/ci/pkg/ci/polaris.go
ivanfetch-fw: That comment was part of the original code and I left it because I wasn't sure if it was useful to someone.

plugins/ci/pkg/ci/helm.go
ivanfetch-fw: That was addressing an error in processing the temporary file name used to write Helm values defined in fairwinds-insights.yaml. This has been separately addressed by another PR from Vitor.

plugins/ci/pkg/ci/helm.go
ivanfetch-fw: I'll update the error message filename when the Helm chart is local (contained in the repo) to use the chart path. I'll keep remote charts using the Helm chart name from fairwinds-insights.yaml, I can't think of a better filename to use in that case.

vitorvezani:
@ivanfetch-fw we might wanna to hold-off the merge until the fix for:
- [FWI-3386](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-3386) [CI Plugin] skip `images.docker` imgs that uses env. variables on CI plugin 

gets in, since we cannot release hotfixes

PR #732, opened 2022-12-06
FWI-3257 - Enable tfsec report by default and gracefully react if no paths were specified

ivanfetch-fw:
## Checklist
* [x] I have signed the CLA

Enable the tfsec report by default, returning gracefully if no terraform paths were specified

Please see the ticket for a deeper description of behavior changes in tfsec, including QA steps.

[Internal Ticket FWI-3257](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-3257)


plugins/ci/pkg/ci/tfsec.go
```diff
@@ -17,37 +17,44 @@ func (ci *CIScan) TerraformEnabled() bool {
...
 		}
 	}
+	if len(reportProperties.Items) == 0 {
+		logrus.Infoln("no Terraform results were returned")
+		return models.ReportInfo{}, false, nil
+	}
 	TFSecVersion, err := commands.Exec("tfsec", "-v")
 	if err != nil {
-		return models.ReportInfo{}, fmt.Errorf("cannot get the version of tfsec: %w", err)
+		return models.ReportInfo{}, false, fmt.Errorf("cannot get the version of tfsec: %w", false, err)
```
vitorvezani, line 41: This looks wrong: `fmt.Errorf("cannot get the version of tfsec: %w", false, err)`

vitorvezani (commented):
Left a comment on an error message
--
Another thing that I thought about... is default value change (false → true) a breaking change in this case? This is not to enforce a major change on the version bump, but just something to wonder about

ivanfetch-fw:
> Another thing that I thought about... is default value change (false → true) a breaking change in this case? This is not to enforce a major change on the version bump, but just something to wonder about

@vitorvezani Yes, this is a breaking change in the case where a user specified `terraform -> paths` but did not intend for those paths to be scanned by relying on the previous default behavior of the `tfsec` report being disabled.

Although this wasn't necessarily your intent, perhaps this *is* a candidate for a major version bump, because we're changing the user experience for this report type. I will update the version to be a major bump, but we can change it.

vitorvezani (approved):

PR #740, opened 2023-01-09
FWI-3386 -skip downloading in-container `images.docker` images that has env. va…

vitorvezani:
This PR fixes #

## Checklist
* [x] I have signed the CLA

Skips downloading in-container `images.docker` images that have env. variables on their names


plugins/ci/pkg/ci/ci.go
```diff
@@ -77,6 +78,7 @@ func NewCIScan() (*CIScan, error) {
 	}
 
 	ci := CIScan{
+		autoScan:       cloneRepo,
```
ivanfetch-fw, line 81: Are auto-scan, and whether the repository is being cloned, this reliably coupled? I am a bit confused by this line.

ivanfetch-fw: Thanks. A comment in the struct for the `cloneRepo` member would help others understand - although non-blocking.

vitorvezani: yep, I've added the comment to the [cloneRepo variable](https://github.com/FairwindsOps/insights-plugins/pull/740/files#diff-044cde4902d1869165353b9fc0e73b91be6154df96d8e5ecdbd18fd7a3ed4326R61) as it is a local var

PR #749, opened 2023-01-25
fwi-3419: prometheus network metrics

ivanfetch-fw:
IMPORTANT: Do not merge this PR, until [this backend PR](https://github.com/FairwindsOps/Insights/pull/3736) has been deployed to production.

## Checklist
* [x] I have signed the CLA

* Add `network receive bytes` and `network transmit bytes` to the metrics collected and submitted to the Insights API.
* Switch from the [controller-utils](https://github.com/FairWindsOps/controller-utils) `GetAllTopControllersSummary` function, to `GetAllTopControllersWithPods`, so that the `Pods` workload struct member will be populated for iteration elsewhere in this plugin. See [this SLack thread](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1677668454068619) for a confirming discussion.

These metrics are represented as totals; cumulative metrics. This code collects extra metrics outside of the specified prometheus range, to use as a baseline while converting these "total" values into delta values. These extra (baseline) metrics are not reflected in the data that is sent to the Insights API.

### Sample Data

Here is a sample single metric for `NetworkTransmit`. Note that the `request` and `limit value` fields are present be use they are part of our existing struct.
```
      {                                                                           
        "Container": "postgres",                                                  
        "PodName": "testapp-845b5f8ffc-tj2j6",                                    
        "ControllerNamespace": "insights-agent",                                  
        "ControllerName": "testapp",                                              
        "ControllerKind": "Deployment",                                           
        "StartTime": "2023-02-16T20:24:00Z",                                      
        "Metric": "NetworkTransmit",                                              
        "Value": 30,                                                              
        "Request": 0,                                                             
        "LimitValue": 0                                                           
      },                                                                          
```


plugins/prometheus/pkg/data/data.go
```diff
@@ -212,3 +268,119 @@ func getOwner(sample *model.SampleStream) Owner {
...
+// each have a value of 3.
+// A metric value of 1.5, will be converted to 3 metrics. The metric for the
+// first container will have value 1.5, and the second and third metrics will
+// each have a value of 0.
+//
+// The specified map of Kubernetes workloads is used to determine the number of
+// containers and their names, for pods referenced in each metric. The map
+// should be
+// keyed on a string `{namespace}/{podName}`.
+func adjustMetricsForMultiContainerPods(metrics model.Matrix, workloadMap map[string]*controller.Workload) model.Matrix {
```
ivanfetch-fw, line 289: This function is a bit long, however I didn't break it out into smaller functions because doing so requires me to pass more values in-and-out of smaller functions, trading this long function for the complexity of more setup and value-passing.

plugins/prometheus/CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
 # Changelog
 
+## 1.1.9
```
rbren, line 3: 🤔 maybe minor bump?

plugins/prometheus/CHANGELOG.md
```diff
@@ -1,10 +1,15 @@
...
+* Add network transmit bytes, and network received bytes, to the metrics submitted to Insights.
+* Add ability to output debug logs using the `LOGRUS_LEVEL` environment variable.
+
 ## 1.1.8
 * update dependencies
 
 ## 1.1.7
 * update alpine and go modules
+>>>>>>> main
```
rbren, line 12: You've got a git merge thing in here

rbren (commented):
Looks great, thanks Ivan! Appreciate the tests

rbren (approved):

plugins/prometheus/pkg/data/data.go
```diff
@@ -171,14 +184,18 @@ func GetMetrics(ctx context.Context, dynamicClient dynamic.Interface, restMapper
...
+	// per-container.
 	client := controller.Client{
-		Context: ctx,
-		Dynamic: dynamicClient,
+		Context:    ctx,
+		Dynamic:    dynamicClient,
 		RESTMapper: restMapper,
 	}
-	workloads, err := client.GetAllTopControllersSummary("")
+	// This used to use client.GetAllTopControllersSummary(), which I believe
```
ivanfetch-fw, line 196: @rbren I wanted to call this out and make sure you weigh in on this change since you updated other projects to use newer versions of our controller-utils. I believe we want to fetch Pods in the prometheus plugin, regardless of the new network metrics. THIs could also be a significant increase of resources pulled from in-cluster though.

rbren: So both of these functions loop through all the pods in the cluster. The one you're using returns a _full_ list of pods, the old one just return (IIRC) a count and a single podspec.

It doesn't seem like you're using the individual pods, so maybe we can revert this?

ivanfetch-fw: We had [a further discussion in Slack](https://fairwindsops.slack.com/archives/CNQBGRGFN/p1677668454068619) - some code here does iterate `workload.Pods`, we will keep the switch to `controller-utils.GetAllTopControllersWithPods`.

PR #755, opened 2023-02-10
Fix up fairwinds-insights.yaml

rbren:
This PR adds images to fairwinds-insights.yaml to ensure they get scanned via Insights

## Checklist
* [x] I have signed the CLA

Scan all images via insights

The image list used to be auto-generated. That doesn't work with autoscan though, so now we need to keep it up-to-date.

Move back to manual scan


sudermanjr:
@rbren Does this still need to get done?

PR #766, opened 2023-03-16
Fix trivy rolling scans when there are many errored images

rbren:
This PR fixes #

## Checklist
* [x] I have signed the CLA

When trivy fails to scan many images, it ends up re-scanning the same images over and over, instead of rolling to new images.

Simplified the logic for keeping images in the report, to make sure it's correct


plugins/trivy/pkg/image/util.go
```diff
@@ -0,0 +1,107 @@
...
+func GetUnmatchingImages(baseImages []models.ImageDetailsWithRefs, toMatch []models.Image, isRecommendation bool) []models.ImageDetailsWithRefs {
+  return getImages(baseImages, toMatch, isRecommendation, false)
+}
+
+func getImages(baseImages []models.ImageDetailsWithRefs, toMatch []models.Image, isRecommendation bool, match bool) []models.ImageDetailsWithRefs {
+	filtered := make([]models.ImageDetailsWithRefs, 0)
+	isMatch := convertImagesToMap(toMatch)
+	isRepoMatch:= imagesRepositoryMap(toMatch)
+	for _, im := range baseImages {
+		fmt.Println("im", im.Name)
```
jdesouza, line 24: fmt.Print

jdesouza (approved):

PR #774, opened 2023-04-14
FWI-3823 - Workloads report should report "requests" being set if "limits" are set, even if "requests" aren's explicit

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA

[Internal Ticket FWI-3823](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-3823)


plugins/workloads/pkg/info.go
```diff
@@ -119,7 +119,12 @@ func formatContainer(container corev1.Container, containerStatus corev1.Containe
 			Memory: container.Resources.Limits.Memory().String(),
 		},
 	}
-
+	if container.Resources.Requests.Cpu().IsZero() && !container.Resources.Limits.Cpu().IsZero() {
+		resources.Requests.CPU = resources.Limits.CPU
+	}
+	if container.Resources.Requests.Memory().IsZero() && !container.Resources.Limits.Memory().IsZero() {
+		resources.Requests.Memory = resources.Limits.Memory
+	}
```
rbren, line 127: I'm actually kinda surprised `Cpu()` and `Memory()` don't return `nil`...but we'd have seen that problem already if so.

PR #775, opened 2023-04-14
Add namespace allowlist to trivy

rbren:
## Checklist
* [x] I have signed the CLA

Add an allowlist in addition to the blocklist. Also change blacklist to blocklist


plugins/trivy/pkg/image/getimages.go
```diff
@@ -17,6 +17,32 @@ import (
...
+var namespaceAllowlist []string
+
+func init() {
+	namespaceBlocklist = strings.Split(os.Getenv("NAMESPACE_BLACKLIST"), ",")
+	namespaceBlocklist = append(strings.Split(os.Getenv("NAMESPACE_BLOCKLIST"), ","), namespaceBlocklist...)
+	namespaceAllowlist = strings.Split(os.Getenv("NAMESPACE_ALLOWLIST"), ",")
+}
+
+func namespaceIsBlocked(ns string) bool {
+	for _, namespace := range namespaceBlocklist {
```
jdesouza, line 30: No need to lower case ns?

plugins/trivy/pkg/image/getimages.go
```diff
@@ -17,6 +17,32 @@ import (
...
+		if ns == strings.TrimSpace(strings.ToLower(namespace)) {
+			return true
+		}
+	}
+	if len(namespaceAllowlist) == 0 {
+		return false
+	}
+	for _, namespace := range namespaceBlocklist {
```
jdesouza, line 38: Should not this be the allowed list?

rbren, line 38:
```suggestion
	for _, namespace := range namespaceAllowlist {
```

plugins/trivy/pkg/image/getimages.go
rbren: Nah that's coming straight from the k8s API, so it'll always be lower

jdesouza (approved):

PR #786, opened 2023-06-01
Update dependencies

jslivka:
This PR updates alpine base image on ci, fw-kube-bench, fw-kubesec, fw-trivy, utils images to address CVEs. Update goldilocks to v4.8.2 and aws-costs to 1.3.0

## Checklist
* [x] I have signed the CLA

Fix CVEs

Update dependencies


rbren (approved):

jslivka:
I've backed out my changes in this PR to introduce a more robust approach to vulnerability scanning for a a couple of reasons:

* It was difficult to reason about when the scan should run
* We push a new image with the commit SHA only if the image configuration changed, and the scan is happening downstream. There isn't a straightforward way to tell if the image simply failed to build previously or if does not exist because it wasn't changed

This now only contains the base image updates to address CVEs

fairwinds-insights.yaml
```diff
@@ -4,23 +4,23 @@ options:
 # run ./scripts/scan-all.sh to regenerate
 images:
   docker:
-    - quay.io/fairwinds/polaris:7.3
+    - quay.io/fairwinds/polaris:7.4
```
vitorvezani, line 7: Are these upgrades supposed to be in this PR?

scripts/scan-all.sh
```diff
@@ -2,7 +2,7 @@
 set -eo pipefail
 
 # Hard-coding four external images we own. Versions taken from insights-agent. Need to find a better solution here.
-images=(quay.io/fairwinds/polaris:7.4 quay.io/fairwinds/nova:v3.6 us-docker.pkg.dev/fairwinds-ops/oss/pluto:v5.16 us-docker.pkg.dev/fairwinds-ops/oss/goldilocks:v4.7)
+images=(quay.io/fairwinds/polaris:7.4 quay.io/fairwinds/nova:v3.6 us-docker.pkg.dev/fairwinds-ops/oss/pluto:v5.16 us-docker.pkg.dev/fairwinds-ops/oss/goldilocks:v4.8)
```
vitorvezani, line 5: is `goldilocks` upgrade supposed to be in this PR?

jslivka: Good question, I think this is just for the security scan. I maybe assumed wrong that we wanted these to be pointed at the latest versions? Should that be addressed somewhere else. This also applies to your comment above

vitorvezani: Yeah, I think it does make sense...

In [fairwinds-insights.yaml](https://github.com/FairwindsOps/insights-plugins/pull/786/files#diff-cd7b940b608e002b123a20a164d7dcd21dfcbe56242d68cc6d378afa5bf944be) we explicitly point out the images that we use, so they can be scanned

vitorvezani (approved):

fairwinds-insights.yaml
```diff
@@ -4,23 +4,23 @@ options:
...
-    - quay.io/fairwinds/right-sizer:0.4.6
-    - quay.io/fairwinds/fw-trivy:0.24.8
-    - quay.io/fairwinds/insights-uploader:0.4.2
-    - quay.io/fairwinds/insights-utils:0.0.3
-    - quay.io/fairwinds/workloads:2.4.4
+    - us-docker.pkg.dev/fairwinds-ops/oss/pluto:v5.16
+    - us-docker.pkg.dev/fairwinds-ops/oss/goldilocks:v4.8
+    - quay.io/fairwinds/insights-admission-controller:1.9.9
+    - quay.io/fairwinds/aws-costs:1.3.0
+    - quay.io/fairwinds/insights-ci:5.0.4
```
vitorvezani, line 13: based on the insights report, looks like `insights-ci:5.0.4` does not yet exist, so you can use `5.0.3` here

jslivka: That was one reason for my aborted attempt to change the security scanning step a bit in this pipeline.  My understanding right now is that `5.0.4` will be released once this PR is merged into `main`, so we might as well have trivy scanning the latest version of this, until we change how security scanning behaves for branches and PRs.

PR #806, opened 2023-07-27
Various fixes and enhacements to kyverno plugin script

jslivka:
This PR fixes #

## Checklist
* [x] I have signed the CLA

## Description

* Fix log message when checking for clusterreports: this was printing the current namespace which is not applicable to cluster-wide reports
* Include policy/clusterpolicy title and description in report: for some reason the category and message are part of `results` in policyreports, but not `title` and `description` 
* bash refactoring
* remove results with a status of `fail` or `skip` so they are marked as fixed in insights

## Example output

I'll attach examples to the YT ticket


jslivka:
`title` and `description` fields are not required by Kyverno, so they can be blank. Not sure how we'd like to handle this on the insights side

As discussed, on the backend we will accept `[efficiency, reliability, security]` as categories. If none of these are provided, we will fall back on `security`

severities will be mapped 1-1: `critical, high, medium, low, none` we will fall back to `none` if what's provided is not one of these

plugins/kyverno/main.sh
```diff
@@ -1,4 +1,32 @@
 #! /bin/bash
+# Additional information on PolicyReport,ClusterPolicyReport: https://kyverno.io/docs/policy-reports/
+
+populate_policy_metadata() {
+  # to avoid overhead, collect only the last (most recent result) from the policyreport
+  report_json=$(kubectl get $KIND $report -n $namespace -o json | jq '.results |= [.[-1]]')
+  policy_name=$(echo $report_json | jq -r '.results[0].policy')
+
+  # use lookup table to minimize control plane load
+  if [[ ! "${policies[*]}" =~ "${policy_name}" ]]; then
```
rbren, line 10: this might give false positives, per [here](https://stackoverflow.com/questions/3685970/check-if-a-bash-array-contains-a-value)

you could probably check sth like `if -z $policy_titles["$policy_name"]`

rbren:
I'd fall back to the policy title being the check ID (i.e. `metadata.name`)

plugins/kyverno/main.sh
jslivka: This is a bit of a pain, "policy_title" and "policy_description" are allowed to be blank, so the only robust way to check is to keep a list of policies we've already retrieved from the k8s API and check that way.

jslivka: That being said, since all of the policy names are Kubernetes object names, they should never have whitespace in them. So I think this will be fine unless i'm missing something. I do want to clean this up a bit: i don't need to redeclare those variables and the function I introduced isn't as readable as just having some code duplication there IMO

rbren (approved):

rbren (approved):

PR #807, opened 2023-08-03
FWI-4446 - Refactor Trivy plugin to use the shared controller-utils package

jslivka:
This PR fixes #

## Checklist
* [x] I have signed the CLA

## Description

At least 1 insights customer has reported that Trivy action items are getting duplicated for cronjob based jobs and should not be. This is most likely due to the fact that we are not using the shared `controller-utils` package here when we should be.

* Fix a bug where Trivy action items are getting duplicated for cronjob based jobs

* Refactor code to make proper use of `controller-utils`
* Updated `plugins/ci` dependencies due to `go mod tidy` error in CI pipeline:

```
go: updates to go.mod needed; to update it:
	go mod tidy
```

[Internal Ticket FWI-4446](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-4446)


plugins/trivy/pkg/image/getimages.go
```diff
@@ -36,100 +34,71 @@ func namespaceIsBlocked(ns string, namespaceBlocklist, namespaceAllowlist []stri
...
+				}
+				owner.Container = containerStatus.Name
+				imgKey := imageName + "/" + imageID
+				if imageOwners[imgKey] == nil {
+					imageOwners[imgKey] = map[models.Resource]struct{}{}
+				}
+				imageOwners[imgKey][owner] = struct{}{}
+				if _, found := keyToImage[imgKey]; found {
+					continue
+				}
```
rbren, line 96: Some of this mapping logic seems like it could probably be simplified now that we've changed the loop.

Another gotcha here is that, now that we're using controller-utils, not every Controller here will have pods to loop through. E.g. we may find a CronJob with no running pods. So we need a fallback image ID/name.

I think this is maybe worth splitting out into a follow-up though--I tried writing out the logic and it's not a trivial change. And the current PR is a strict improvement over how we're doing things now.

jslivka: Is doing the fallback logic (e.g. if a CronJob doesn't have an active pod scheduled) worth doing in this PR? Or is that also a bear along with the mapping logic for `imageOwners`?

PR #810, opened 2023-08-15
FWI-4713 - Avoid filename collisions in Falco scan for Insights

jslivka:
This PR fixes Internal Ticket FWI-4713

```
time="2023-08-15T16:46:12Z" level=error msg="Error while aggregating data: invalid character '1' after top-level value"
```

The filenames for falco reports stored in `/output` were using unix epoch in seconds. We believe this allows for collisions, thus a situation where two structures are being written to the same file within the same timeframe (1 second). This would result in malformed JSON like we observed in our staging environment:

```
/opt/app $ cat /output/1692039074.json && echo
{"PodName":"trivy-28200645-zmmfj","Container":"insights-uploader","ControllerNamespace":"fw-telemetry","ControllerName":"trivy","ControllerKind":"CronJob","uuid":"d18b95de-9982-4cba-90f3-8de215aa4b1a","output":"18:51:14.837682894: Notice Unexpected setuid call by non-sudo, non-root program (user=\u003cNA\u003e user_loginuid=-1 cur_uid=1000 parent=uploader.sh command=sleep 1 pid=1836280 uid=\u003cNA\u003e container_id=bdde7d7beef5 image=quay.io/fairwinds/insights-uploader) k8s.ns=fw-telemetry k8s.pod=trivy-28200645-zmmfj container=bdde7d7beef5","priority":"Notice","rule":"Non sudo setuid","time":"2023-08-14T18:51:14.837682894Z","output_fields":{"container.id":"bdde7d7beef5","container.image.repository":"quay.io/fairwinds/insights-uploader","evt.arg.uid":"\u003cNA\u003e","evt.time":1692039074837683000,"k8s.ns.name":"fw-telemetry","k8s.pod.name":"trivy-28200645-zmmfj","proc.cmdline":"sleep 1","proc.pid":1836280,"proc.pname":"uploader.sh","user.loginuid":-1,"user.name":"\u003cNA\u003e","user.uid":1000}}1000}}
```

To provide further guarantee of filename uniqueness beyond nanoseconds, I am appending a number at the end. The overhead of `os.Stat` calls should be quite low and this will also preserve ordering in those rare cases.

I've also allowed `Aggregate24hrsData` to skip files it cannot read or unmarshal to JSON, so this will behave in a robust manner with the assumption that malformed json will remain on disk.

Test coverage is not-yet comprehensive, but I've added unit tests for the new utility functions. To better support this, I am wrapping OS filesystem calls so they can be mocked, using [afero](https://github.com/spf13/afero)

Filenames look like this now (Unix epoch to ns):

```
drwxrwsr-x    2 root     insights   2097152 Aug 16 15:26 .
drwxr-xr-x    1 root     root          4096 Aug 16 15:30 ..
-rw-r--r--    1 insights insights      1261 Aug 16 15:31 1692199867765764885.json
-rw-r--r--    1 insights insights      1653 Aug 16 15:31 1692199867665785576.json
/output $
```

## Checklist
* [x] I have signed the CLA

[Internal Ticket FWI-4713](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-4713)


plugins/falco-agent/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.2.9
```
rbren, line 3: This seems like probably worth a minor version bump

plugins/falco-agent/pkg/data/data.go
```diff
@@ -43,24 +44,32 @@ func readDataFromFile(fileName string) (payload FalcoOutput, err error) {
...
 	for _, file := range tmpfiles {
-		if file.Mode().IsRegular() {
+		if file.Type().Perm().IsRegular() {
 			filename := filepath.Join(dir, file.Name())
-			if isLessThan24hrs(file.ModTime()) {
+			fileStat, err = fsWrapper.Stat(filename)
+			if err != nil {
+				return
+			}
+			if isLessThan24hrs(fileStat.ModTime()) {
```
rbren, line 62: Were we not using the timestamp in the filename at all?

rbren (approved):

PR #811, opened 2023-08-21
FWI-4541 [admission] Add 'Fairwinds Insights' indicator to Admission Controller response

jslivka:
This PR fixes #

## Checklist
* [x] I have signed the CLA

Admission Controller feedback in the terminal should display "Fairwinds Insights" like this:

```

Warning: [Fairwinds Insights] Filesystem should not be read only
Warning: [Fairwinds Insights] [Blocked] Container should not run as root
```

I've included example output in the internal ticket


jslivka:
Mixed on how a semver version bump should go for something like this. It's quite a small change, but we are changing the output. Should this be a minor or patch change?

rbren:
Minor seems reasonable. Thanks John!

rbren (approved):

PR #815, opened 2023-08-31
FWI-4897 - Automatically rebuild insights-plugins/main on a regular interval

jslivka:
This PR fixes #

reference: https://circleci.com/docs/migrate-scheduled-workflows-to-scheduled-pipelines/

## Checklist
* [x] I have signed the CLA

This adds a schedule to circleci that instructs it to automatically rebuild images on `main` every Tuesday

[Internal Ticket FWI-4897](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-4897)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -325,3 +325,15 @@ workflows:
             branches:
               only:
                 - main
+
+  rebuild:
+    triggers:
+      - schedule:
+          # At 00:00 every day monday-friday
+          cron: "0 0 * * 1-5"
+          filters:
+            branches:
+              only:
+                - main
+    jobs:
+      - build_and_push_plugins


.circleci/config.yml
```diff
@@ -325,3 +325,15 @@ workflows:
             branches:
               only:
                 - main
+
+  rebuild:
+    triggers:
+      - schedule:
+          # At 00:00 on Tuesday
+          cron: "0 0 * * TUE"
```
rbren, line 333: Let's just do it everyday--it's not that expensive 😄

rbren: or maybe `MON-FRI`?

jslivka: Changed to:

```
# At 00:00 every day monday-friday
          cron: "0 0 * * MON-FRI"
```

jslivka: Edited, think it needs to be `1-5` instead of `MON-FRI` here for circleci

rbren (approved):
Thanks John!

PR #817, opened 2023-09-08
FWI-4945 - Migrate from go-funk library to lo in insights-plugins repo

jslivka:
This PR fixes #

## Checklist
* [x] I have signed the CLA

## Description

For the sake of consistency and readability

Some of the plugins are still using go-funk, refactor to use https://github.com/samber/lo instead

plugins:

    admission
    opa
    right-sizer
    workloads

remove: github.com/thoas/go-funk 
and use github.com/samber/lo

Consistency within `insights` code when using functional programming libraries

Refactored code to use `github.com/samber/lo` instead of `github.com/thoas/go-funk `, where the former utilizes go generics

[Internal Ticket FWI-4945](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-4945)


plugins/right-sizer/src/controller/controller.go
```diff
@@ -133,7 +133,7 @@ func NewController(stop chan struct{}, kubeClientResources util.KubeClientResour
 		o(cfg)
 	}
 
-	missingAllowedNamespaces := funk.RightJoinString(cfg.allowedNamespaces, cfg.allowedUpdateNamespaces)
+	_, missingAllowedNamespaces := lo.Difference(cfg.allowedNamespaces, cfg.allowedUpdateNamespaces)
```
vitorvezani, line 136: I was a bit worried about this one, but they look functionally the same https://go.dev/play/p/dyaPZ-4Pgv2

vitorvezani (approved):
Thanks, John! Not sure if we need a minor release for these changes, but I have no strong opinions.

jslivka:
> Thanks, John! Not sure if we need a minor release for these changes, but I have no strong opinions.

I actually googled this, sounds like it could go either way: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api -- this falls more under "adding new functionality" I think since we are supporting generics, so I went with a minor bump

vitorvezani:
Ohh, yea, makes sense! :shipit:

PR #855, opened 2023-12-14
FWI-5436 - insights-plugins: Add files that were modified to scan response

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA

[Internal Ticket FWI-5436](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-5436)


plugins/ci/pkg/ci/git.go
```diff
@@ -87,9 +89,27 @@ func getGitInfo(cmdExecutor cmdInDirExecutor, ciRunner models.CIRunnerVal, baseR
 			logrus.Warnf("Unable to get GIT branch name: %v", err)
 			gitCommandFail = true
 		}
+		refHash, err := cmdExecutor(baseRepoPath, exec.Command("git", "show-branch", "--merge-base"), "ref commit hash")
```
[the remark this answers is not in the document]
jdesouza: So you saying if the MASTER_HASH is set we should use it instead of trying to get the ref branch from that branch?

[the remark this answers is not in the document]
jdesouza: Refactoring done

plugins/ci/pkg/ci/git.go
```diff
@@ -87,9 +89,30 @@ func getGitInfo(cmdExecutor cmdInDirExecutor, ciRunner models.CIRunnerVal, baseR
 			logrus.Warnf("Unable to get GIT branch name: %v", err)
 			gitCommandFail = true
 		}
+		refHash := os.Getenv("MASTER_HASH")
```
rbren, line 92: Can we resuse `getGitInfo` here?

https://github.com/FairwindsOps/insights-plugins/blob/4db11ca5df2c1e80794cd7d15178058dfd764c76/plugins/ci/pkg/ci/git.go#L30

rbren: Maybe we should add `modified files` to `getGitInfo`

jdesouza: Can't, I tried to reuse it but it gives me different result than I was expecting

rbren: hmm if it gives a different result, then it's going to be inconsistent with how we're doing scans

jdesouza: Reverted my new code in favor of the current one

plugins/ci/pkg/ci/ci.go
```diff
@@ -309,12 +311,23 @@ func (ci *CIScan) sendResults(reports []*models.ReportInfo) (*models.ScanResults
...
+		if err != nil {
+			logrus.Fatalf("Unable to create form for %s: %v", "files_modified", err)
+		}
+		_, err = fw.Write([]byte(strings.Join(repoDetails.filesModified, "\n")))
+		if err != nil {
+			logrus.Fatalf("Unable to write file for %s: %v", "files_modified", err)
+		}
+	}
+
+	w.Close()
```
vitorvezani, line 330: [opt] move `defer w.Close()` to line 298

vitorvezani (approved):

plugins/ci/pkg/ci/ci.go
jdesouza: Still writing to it at this point

vitorvezani: That is what the `defer` would help.
The way it is set, if the program closes by any of `logrus.Fatalf` the writer would never be closed.

PR #862, opened 2023-12-22
FWI-5471 - Fix CI image owner matching and add multiple owners support

vitorvezani:
This PR fixes internal FWI-5471

## Checklist
* [X] I have signed the CLA

 - Fix image owner matching
 - Add multiple owners support

[Internal Ticket FWI-5471](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-5471)


plugins/ci/pkg/ci/trivy.go
```diff
@@ -214,10 +214,23 @@ func mergeImages(folderPath string, dockerImages []trivymodels.DockerImage, mani
...
+			// fallback - match by name
+			if imageName, ok := filenameToImageName[filename]; ok {
+				if im.Name == imageName {
+					if im.PullRef == "" {
+						im.PullRef = filename // set pullRef if not already set
+					}
+					image = &im
+					break
+				}
+			}
```
rbren, line 233: 🤔 setting `PullRef` to `filename` seems odd--though it looks like we were already doing that above...

rbren (approved):

plugins/ci/pkg/ci/trivy.go
vitorvezani: In all my tests `PullRef` was blank, this is kinda of a reverse-lookup using the filename to imagemap, while preserving the original logic

But yea, kinda odd indeed... I will give it a deeper look.

vitorvezani: I was mutating a copy of the slice, that's why `PullRef` was always `""` - fixed that, now we don't need this fallback behavior - thanks for noticing it 🕺

PR #889, opened 2024-04-03
FWI-5959 - Update prometheus reports to bring nodes idle information

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA

[Internal Ticket FWI-5959](https://fairwinds.myjetbrains.com/youtrack/issue/FWI-5959)


plugins/prometheus/pkg/data/data.go
```diff
@@ -92,6 +91,28 @@ func getController(workloads []controller.Workload, podName, namespace string) (
...
+	}
+	logrus.Infof("Found %d metrics for idle CPU", len(idleCPU))
+	idleMemory, err := getNodesIdleMemory(ctx, api, r)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Infof("Found %d metrics for idle Memory", len(idleMemory))
+	response := &NodesMetrics{}
+	if len(idleCPU) > 0 {
+		response.IdleCPU = idleCPU[0].Values[0].Value
```
vitorvezani, line 108: is `Values[0]` always there or requires a length check?

jdesouza: It should be but better adding a check

vitorvezani (approved):

PR #890, opened 2024-04-03
Accomodate the change in Kyverno report aggregation in v1.11+

jslivka:
This PR fixes #FWI-5963

## Checklist
* [x] I have signed the CLA

## Description
Kyverno changed how they name PolicyReport objects. It used to be [cpol]-$POLICY_NAME e.g. cpol-require-labels . Now the name is The uid value, e.g.13627ecf-3603-4794-98e2-5e15be887123The api version remains: [wgpolicyk8s.io/v1alpha2](http://wgpolicyk8s.io/v1alpha2) after this change

verified this change works for the following versions

```
Chart version: 3.0.9
Kyverno version: v1.10.7

Chart version: 3.1.4
Kyverno version: v1.11.4

with `insights-plugins/kyverno` FWI-5963/jbs (proposed v0.2.0)
```

Information on the change in the release notes:

https://github.com/kyverno/kyverno/releases/tag/v1.11.0

> Policy Reports are now created on a per-resource basis and using a UID as the name rather than the previous behavior of per-policy. This may be a breaking change if you relied upon either of these attributes in previous versions. This change has the benefit of putting less pressure on the Kubernetes API server and less storage cost on etcd.

The primary change was in how we enumerate policyreports. Prior to this change, we list out each policyreport (which corresponds to each policy) and then take the `results` section with `pass` and `skip` items filtered out. In this implementation, We take all policyreports, filter for `fail` `warn` or `error` (since now that the reports are per resource, we have a lot of results in each namespace even if they are all passing), and then we proceed with the `results` section with `pass` and `skip` items filtered out, same as before.


plugins/kyverno/main.sh
```diff
@@ -57,44 +76,13 @@ for namespace in "${namespaces[@]}"; do
...
-    policy_description=${policy_descriptions[$policy_name]}
-    jq --arg title "$policy_title" --arg description "$policy_description" '. += {policyTitle: $title, policyDescription: $description}' < $report_file | sponge $report_file
-  fi
+    if [[ "$kind" == "PolicyReport" ]]; then
+    	jq --slurpfile report_json "$report_file" '.policyReports += $report_json' < $list_file | sponge $list_file
+	fi
 
-  jq --slurpfile report_json "$report_file" '.clusterPolicyReports += $report_json' < $list_file | sponge $list_file
+	if [[ "$kind" == "ClusterPolicyReport" ]]; then
+    	jq --slurpfile report_json "$report_file" '.clusterPolicyReports += $report_json' < $list_file | sponge $list_file
```
sudermanjr, line 84: nit: indentation

plugins/kyverno/main.sh
```diff
@@ -57,44 +76,13 @@ for namespace in "${namespaces[@]}"; do
...
-    policy_titles["$policy_name"]=$policy_title
-    policy_descriptions["$policy_name"]=$policy_description
-  else
-    policy_title=${policy_titles[$policy_name]}
-    policy_description=${policy_descriptions[$policy_name]}
-    jq --arg title "$policy_title" --arg description "$policy_description" '. += {policyTitle: $title, policyDescription: $description}' < $report_file | sponge $report_file
-  fi
+    if [[ "$kind" == "PolicyReport" ]]; then
+    	jq --slurpfile report_json "$report_file" '.policyReports += $report_json' < $list_file | sponge $list_file
+	fi
```
sudermanjr, line 81: nit: indentation

sudermanjr (approved):
indentation across the board is a bit funky, not a blocker

jdesouza (approved):

PR #894, opened 2024-04-18
Added aggregation to GCP billing report

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA


plugins/cloud-costs/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.2.2
```
rbren, line 3: should we make this a minor bump just in case?

jdesouza: We could but 0.2.1 has a bug - we would need to patch it as well, so not sure if it's worthy

rbren (approved):

PR #926, opened 2024-05-09
Added google access token to prometheus plugin

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA


plugins/prometheus/cmd/prometheus-collector/main.go
```diff
@@ -31,12 +33,27 @@ import (
 )
 
 const outputFile = "/output/prometheus-metrics.json"
+const monitoringReadScope = "https://www.googleapis.com/auth/monitoring.read"
+const monitoringGoogleApis = "monitoring.googleapis.com"
 
 func main() {
 	setLogLevel()
 	address := os.Getenv("PROMETHEUS_ADDRESS")
```
vitorvezani, line 41: should we fail fast if `address` is empty?

jdesouza: added, although it fails fast if url is missing

vitorvezani: yeah, this way we can provide a better message to the end-user… 
[opt] I wonder if `PROMETHEUS_ADDRESS must be set` would be better as `prometheus-metrics.address` only makes sense from whom is using this via helm(?)

vitorvezani (approved):

PR #974, opened 2024-10-21
2 add gcloud trivy/vv/insights 410

vitorvezani:
This PR fixes #

## Checklist
* [X] I have signed the CLA

Use `gcloud` auth in case it's available - this approach will use `gloud` auth command to set up dockerconfig


vitorvezani:
`gcloud auth configure-docker` requires `docker` installed, and we are trying to avoid that -- so we are going to implement using tokens for now

PR #983, opened 2024-12-06
INSIGHTS-527 -  Support multiple rules that return values in OPA

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA

[Internal Ticket INSIGHTS-527](https://fairwinds.myjetbrains.com/youtrack/issue/INSIGHTS-527)


plugins/opa/pkg/opa/opa_test.go
```diff
@@ -100,6 +100,27 @@ requestinsightsinfo[description] {
 }
 `
 
+const maxReplicas = `
+package fairwinds
+
+foo := {"s": "foo"}
+
+envMaxReplicasDeployments {
```
vitorvezani, line 108: please, also add a test using
```envMaxReplicasDeployments[actionItem] {
    ...
}```

which is usually how policies are defined

jdesouza: done

vitorvezani (approved):

vitorvezani:
We need to do similar in [the cli](https://github.com/FairwindsOps/insights-cli/blob/e7ab573d96934142eee24ac55dfdf707ace3fc5a/pkg/opavalidation/utils.go#L24)

PR #1031, opened 2025-06-06
INS-1116: Updated trivy and cloud-costs to python image

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA


plugins/cloud-costs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.22
+FROM python:alpine3.22
```
vitorvezani, line 1: [opt] use `python:3-alpine3.22` instead, if python 4 is ever released, might break 😬

PR #1172, opened 2026-04-20
INS-2229: Add support to status.containerStatuses[].resources (runtime-applied) in workloads report

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA


plugins/workloads/results.schema
```diff
@@ -362,11 +362,51 @@
                             "Memory",
                             "CPU"
                           ]
+                        },
+                        "SpecAppliedConvergedCount": {
```
vitorvezani, line 366: if `SpecAppliedConvergedCount == len(SpecAppliedSkewPods)`, I think we can remove it

jdesouza: You mean if we have no in place resize? or if we have only in-place changes?

plugins/workloads/pkg/actual.go
```diff
@@ -0,0 +1,176 @@
...
+		out.Requests.Memory = out.Limits.Memory
+	}
+	return &out
+}
+
+func qtyString(rl corev1.ResourceList, name corev1.ResourceName) string {
+	if q, ok := rl[name]; ok {
+		return (&q).String()
+	}
+	z := resource.MustParse("0")
```
vitorvezani: Missing resources are set to 0, maybe that is ok, maybe it isn't, just pointing it out

plugins/workloads/results.schema
vitorvezani: nvm. I though `SpecAppliedConvergedCount == len(SpecAppliedSkewPods)` was always true but dosnt seem to be the case

vitorvezani (approved):

plugins/workloads/pkg/actual.go
jdesouza: done

PR #1189, opened 2026-05-01
INS-2296: Renovate for insights-plugins is missing some updates

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA


plugins/kube-bench/Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apk --no-cache add bash ca-certificates curl \
 
 RUN rm /usr/local/bin/kubectl
 ENV kubectlVersion=1.35.3
```
vitorvezani: I'm not sure but maybe having 
```
ENV kubectlVersion=v1.35.3
RUN curl -LO https://dl.k8s.io/release/$kubectlVersion/bin/${TARGETOS}/${TARGETARCH}/kubectl
```
would work best for renovate

jdesouza: that's kind of weird because in places we do have the v

jdesouza: ah I see now

vitorvezani (approved):

PR #1279, opened 2026-07-20
INS-2607: Add nodes placements to plugin

jdesouza:
This PR fixes #

## Checklist
* [x] I have signed the CLA


plugins/workloads/pkg/info.go
```diff
@@ -44,6 +46,8 @@ type ControllerResult struct {
 	Containers     []ContainerResult
 	// VolumeClaims lists PersistentVolumeClaim refs from the pod template and running pods.
 	VolumeClaims []VolumeClaimRef `json:",omitempty"`
+	// NodeNames are unique nodes where Running pods for this controller are scheduled.
+	NodeNames []string `json:",omitempty"`
```
vitorvezani, line 50: any change we need to emit extra nodes info in the future? if so, I would use
```
nodes: [{
 name: "node-a"
}]
``` 
so we can easily expand later

vitorvezani (approved):

